PR #41, opened 2026-06-05
feat: add support for Expo SDK 56

equationalapplications:
## Summary
- Updated root `package.json` and `packages/expo/package.json` for Expo SDK 56 support
- Regenerated `pnpm-lock.yaml` to reflect dependency changes

## Changes
- Modified `package.json` (root)
- Modified `packages/expo/package.json`
- Updated `pnpm-lock.yaml`

## Notes
This commit follows conventional commit format to trigger semantic-release on merge to main.


equationalapplications:
## PR 41 Review — Results Summary

All 9 findings verified. **7 fixed, 1 skipped as not valid, 1 already resolved.**

| # | Finding | Verdict | Action |
|---|---------|---------|--------|
| **1** | Doc: wrong `SQLiteAdapter` import from `@equationalapplications/core-llm-wiki` | ✅ Valid | Fixed: replaced with locally-defined interface + import from `./sqljs-adapter` |
| **2** | Doc: `locateFile` returns `/wasm/${file}` conflicts with `/sql-wasm.wasm` | ✅ Valid | Fixed: updated to `/sql-wasm.wasm` with comment explaining build-time copy |
| **3** | Doc: XOR/base64 presented as viable "encryption" | ✅ Valid | Replaced: removed entire XOR snippet, replaced with security warning + sessionStorage + Web Crypto recommendation |
| **4** | App.tsx: stale `history` passed to `chatWithMemory` | ❌ **Not valid** | `chatWithMemory` receives `history` (prior context) + `userMessage` separately and assembles `[...history, {user}]`. Passing `nextHistory` would double-add the user message. |
| **5** | index.css: deprecated `word-break: break-word` | ✅ Valid | Fixed: replaced with `overflow-wrap: anywhere; word-break: normal` |
| **6** | function-caller.ts: tool lookup excludes `'core'` scope | ✅ Valid | Fixed: added `t.scope === 'core'` fallback in `tools.find` predicate |
| **7** | tool-executor.ts: `search_memory` stubs, unknown tools fake success | ✅ Valid | Fixed: `search_memory` queries wiki via `setToolWiki` wiring; unknown tools throw `Error` |
| **8** | wiki-service.ts: `persist` returns empty `Uint8Array` on failure | ✅ Valid | Fixed: throws descriptive `Error` when adapter/export is missing |
| **9** | sqljs-adapter.ts: hardcoded `/sql-wasm.wasm` breaks subpath deploys | ✅ Valid | Fixed: resolved via `import.meta.env.BASE_URL` |

### Additional wiring

- Added `setToolWiki(service)` call in App.tsx so `tool-executor` has access to wiki for real `search_memory` queries.

### Validation

- ✅ All 105 tests pass across all packages
- ✅ Zero TypeScript errors in all 5 modified source files

equationalapplications:
Changes made
1. sqljs-adapter.ts — WASM statement leak fix
Wrapped stmt.bind()/stmt.step() in try/finally across runAsync, getAllAsync, and getFirstAsync so stmt.free() always executes even if an intermediate call throws.

2. function-caller.ts — HTTP error handling for Gemini API calls
Added response.ok checks before response.json() on both the initial Gemini call and the follow-up (tool response) call. Non-2xx responses now throw with the status code and response body excerpt instead of silently parsing into bad data.

PR #60, opened 2026-07-09
fix(core): serialize SQLite transactions on the shared connection

equationalapplications:
## Summary
- Wrap the `SQLiteAdapter` in a promise-chain mutex (`withSerializedTransactions`) once in the `WikiMemory` constructor so all 11 transaction call sites serialize with no per-site change. Fixes the concurrent-write cascade: `cannot start a transaction within a transaction` → `cannot rollback - no transaction is active` → wedged connection.
- Inject a reentrancy guard onto every `tx` handle so a nested `withTransactionAsync` throws synchronously instead of deadlocking; a >10s queue wait logs a deadlock warning.
- Wrap escaping driver errors in a typed `WikiTransactionError` (with lifted `sqliteErrorCode`); domain errors pass through unwrapped with `instanceof` intact.
- Guard rollback in the test-helper adapter so a failed `BEGIN` never masks the original error.
- Document the concurrency contract + `WikiTransactionError` in core/expo READMEs and JSDoc.

Uses prototype delegation (`Object.create`) rather than object-spread in the wrapper so class-instance adapters keep their prototype methods.

## Test Plan
- [x] `packages/core` — 709 tests pass (new `transactionSerialization.test.ts` + all pre-existing)
- [x] Concurrent-write regression fails without the fix (`cannot start a transaction within a transaction`), passes with it
- [x] `packages/expo` — build + 10 tests pass
- [x] `packages/integration` — 37 tests pass
- [x] `packages/core` build type-checks clean

### DIFF
--- a/docs/superpowers/specs/2026-07-09-transaction-serialization-spec.md
+++ b/docs/superpowers/specs/2026-07-09-transaction-serialization-spec.md
@@ -11,7 +11,7 @@
 Production incident (host app, web, 2026-07-09): concurrent wiki operations on a single
 shared SQLite connection produced:
 
-```
+```text
 Error code 1: cannot start a transaction within a transaction
 Error code 1: cannot rollback - no transaction is active
 ```
@@ -135,12 +135,10 @@ function guardReentrancy(tx: SQLiteAdapter): SQLiteAdapter {
 The queue holds a single promise chain; each resolved link is dropped as the
 tail advances, so V8/JSC reclaims settled transactions over a long-running app.
 
-> **Adapter contract — plain objects only.** The wrapper (and `guardReentrancy`)
-> use object spread (`{ ...db }`), which copies **own enumerable** properties only.
-> The expo adapter and the core test helper are plain object literals, so this is
-> correct. A future adapter authored as a **class instance** would lose its
-> prototype methods and `this` binding through the spread — adapters MUST be plain
-> objects, or the wrapper must be rewritten to delegate each method explicitly.
+> **Adapter contract.** The wrapper (and `guardReentrancy`) use prototype
+> delegation (`Object.create(...)`), not object spread, so inherited methods and
+> `this` binding are preserved — a class-instance adapter keeps its prototype
+> methods through the wrap.
 >
 > **Expo `tx` aliasing.** `packages/expo/src/adapter.ts` passes the outer adapter
 > object itself as the `tx` argument (it wraps `expo-sqlite`'s own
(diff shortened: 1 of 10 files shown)


equationalapplications:
## /fix-pr follow-up

**Commit:** `fc8dddb`

### Review resolution
- **Fenced code blocks missing language tags** — Fixed: added `text` language tags to both bare fences in the spec (files: `docs/superpowers/specs/2026-07-09-transaction-serialization-spec.md`)
- **Stale "plain objects only" adapter contract note** — Fixed: rewrote the note to describe the actual `Object.create` prototype-delegation approach, which preserves inherited methods for class-based adapters (files: `docs/superpowers/specs/2026-07-09-transaction-serialization-spec.md`)
- **Test asserting `.rejects.not.toThrow` on a resolving promise** — Fixed: rewrote the masking-error test to capture the error from the single `withTransactionAsync` rejection and assert on its message directly, instead of issuing a second independent call (files: `packages/core/__tests__/transactionSerialization.test.ts`)
- **Expo numeric SQLite codes not normalized to symbolic names** — Fixed: added a SQLite primary-result-code → symbolic-name table (`1` → `SQLITE_ERROR`, `5` → `SQLITE_BUSY`, etc., falling back to `SQLITE_<n>` for unmapped codes) used by both `extractSqliteCode` (serializedAdapter.ts) and the local copy in `types.ts`, so busy/error checks are consistent across drivers. Updated the existing test expectation (`SQLITE_1` → `SQLITE_ERROR`) and added coverage for a mapped and an unmapped code.

### Verification
- typecheck — pass (`pnpm run typecheck` in `packages/core`)
- lint — N/A (no lint script in `packages/core`); markdownlint MD040 (fenced-code-language) now clean on the spec file — pre-existing MD013 line-length warnings in that file are unrelated to this review and left as-is
- tests — pass, 709/709 (`pnpm run test` in `packages/core`)

equationalapplications:
## /fix-pr follow-up

**Commit:** `0a02f9ac3a5e382e67fe6cffe0ac2415782160c7`

### Review resolution
- **Duplicated SQLITE_RESULT_CODE_NAMES / extractSqliteCode (Copilot + CodeRabbit nitpick)** — Fixed: extracted into new dependency-free `packages/core/src/db/sqliteCodes.ts`, imported by both `db/serializedAdapter.ts` and `types.ts` (`WikiTransactionError`). No more manual "keep in sync" duplication.
- **`Object.create` breaks `this` receiver for class-based adapters (#private fields / instanceof)** — Fixed: replaced with a `Proxy`-based `withOverrides` helper in `serializedAdapter.ts` that binds delegated methods to the original adapter instance, used by both `guardReentrancy` and `withSerializedTransactions`.
- **`withTransactionAsync` JSDoc implies serialization is guaranteed by the adapter itself** — Fixed: reworded in `types.ts` to clarify serialization is provided by `WikiMemory`'s `withSerializedTransactions` wrapper, not a bare adapter implementation.

### Verification
- typecheck — pass (`pnpm typecheck`, all 9 workspace packages)
- lint — N/A (no lint script in this repo)
- tests — pass (`pnpm test`: core 709/709, core-llm-tools 51/51, okf 81/81, react 70/70, expo 10/10, prisma-outbox 15/15, integration 35/35)

Issue #65, opened 2026-08-04
doRunOntologyBackfill also fails on large per-entity backlogs (same defect class as #63)


equationalapplications:
## Summary

`MaintenanceService.doRunOntologyBackfill` fails repeatably on a large per-entity untyped-fact backlog, in the same way `doRunHeal` does (#63): its output is not bounded to fit within a model's max-output-tokens budget, so once the backlog is large enough the call fails — and, unlike a transient error, it fails again on every retry, because retrying resends the same oversized request.

This reproduced 3 times in a row against the same backlog (168-document corpus, `business` entity, 1177 untyped facts), via Lambda's automatic async retry (3 attempts total, same invocation). Two distinct failure shapes were observed across those attempts:

1. `SyntaxError: No JSON object/array found in LLM response` — the model's output apparently wasn't parseable JSON at all (possibly truncated somewhere that broke the JSON grammar, or a non-JSON preamble).
2. `Error: Model response truncated at the 9999-token limit` — an explicit, unambiguous truncation.

Both point at the same underlying cause: the backfill call for this entity needs more output budget than is available, and which failure mode surfaces appears to depend on exactly where truncation lands relative to the JSON structure.

## Context: this is not solvable by raising max tokens further

The consuming project (`aws-cloud-agent`) had already hit this same class of problem in `doRunHeal` (#63) and raised its tier's `maxTokens` as a mitigation — first to 16384 (rejected outright: Amazon Nova Pro enforces a hard ceiling, `"The maximum tokens you requested exceeds the model limit of 10000"`), then to 9999 (the actual ceiling). `doRunOntologyBackfill` still fails at 9999. There is no more headroom available on this model to raise into — the fix has to be on the output-shape side, not the token-budget side.

## Where

`packages/core/src/services/MaintenanceService.ts`, `doRunOntologyBackfill` (around line 622, per the stack trace: `MaintenanceService.doRunOntologyBackfill (.../MaintenanceService.ts:622:...)`, called from the run loop that also produces the `passes`/`converged` reporting recently added downstream).

Given `ontologyBackfillBatchSize` already exists as a configurable option (consumed as `runOntologyBackfill(entityId, { batchSize })`), the batching mechanism may already exist but not be sized correctly for a call that also has to describe/reason about a large fact set per pass — or the per-pass output (not just the input batch) needs its own bound that isn't currently enforced.

## Observed data

| | |
|---|---|
[...]

Issue #64, opened 2026-08-04
Uncaught TypeError in minisearch's performVacuuming crashes the process during concurrent ingest


equationalapplications:
## Summary

An uncaught `TypeError` inside `minisearch`'s internal auto-vacuum (`MiniSearch.performVacuuming` → `TreeIterator.next` → `TreeIterator.dive`) crashed the Node process outright during a live ingest run. It happened on the **first attempt** of two separate ingest invocations, at different points in each run (after document 9 in one, document 9 again in the other — early, but not at a fixed document count), and did not happen on either invocation's retry.

Because it's an uncaught exception outside any try/catch this package's callers control, it kills the whole process. On Lambda this only "worked" because the invocations were asynchronous and got auto-retried; a synchronous caller has no equivalent safety net.

## Where

`node_modules/minisearch` v7.2.0, internal to the package — not called directly by `SearchService`. The stack trace:

```
/var/task/node_modules/minisearch/src/SearchableMap/TreeIterator.ts:62
  this._path.push({ node: child, keys: Array.from(child.keys()) })
                                              ^
TypeError: Cannot read properties of undefined (reading 'keys')
    at TreeIterator.dive (.../minisearch/src/SearchableMap/TreeIterator.ts:62:59)
    at TreeIterator.next (.../minisearch/src/SearchableMap/TreeIterator.ts:51:24)
    at MiniSearch.performVacuuming (.../minisearch/src/MiniSearch.ts:1106:35)
```

`child` is `undefined` where the iterator expects a node with a `.keys()` method — consistent with a tree node being mutated (removed) concurrently with, or between, the vacuum's traversal steps.

## How it was triggered

`SearchService` (`packages/core/src/services/SearchService.ts`) drives `minisearch` via `discard()`/`add()`/`replace()` during `sync()`, called once per ingested document during a real ingest run against a 168-document corpus at `documentConcurrency: 3`. MiniSearch's `autoVacuum` is on by default and fires internally on its own schedule (a discard-count threshold), asynchronously with respect to the caller.

Suspect: with concurrent document ingestion (§4.3 of the consuming project's design — three documents processing in parallel, each syncing the search index as it completes), `sync()` calls for different documents could interleave, and an internal vacuum pass triggered mid-interleaving could observe a partially-mutated tree — exactly the shape of `undefined` where a live node was expected. Not confirmed; offered as the most consistent theory of "first attempt only, position varies, never reproduced on retry."

## Reproduction

[...]

PR #66, opened 2026-08-04
Bound maintenance output, fix search rebuild race, clean up dependencies

### LINKED ISSUE #63: doRunHeal sends every entity fact in one un-paginated prompt; breaks permanently past a corpus threshold
## Summary

`MaintenanceService.doRunHeal` builds a single prompt containing **every fact for an entity**, with no pagination or chunking, and expects the model to answer with a single JSON response. Prompt size therefore grows linearly with corpus size, and past a certain corpus the call fails outright — permanently, on every subsequent run, with no way to make progress.

This is not a hypothetical: it reproduced on a real corpus and currently blocks `heal` entirely for that dataset.

## Where

`packages/core/src/services/MaintenanceService.ts`, in `doRunHeal` (~line 468):

```ts
const allFactsRows = await this.entryRepo.findAllByEntityId(entityId);
const allTasks     = await this.taskRepo.findAllPending([entityId]);
const recentEvents = await this.eventRepo.getRecent(entityId, 20);

const healCandidates = allFactsRows.filter(f => f.source_type !== 'immutable_document');
const documentAnchors = allFactsRows
  .filter(f => f.source_type === 'immutable_document')
  .map(({ id, title, source_ref }) => ({ id, title, source_ref }));
```

Both derived arrays flow unbounded into `buildHealPrompt`, then into one `generateText` call.

Note the inconsistency with its sibling: `runLibrarian` **is** bounded, at lines 307 and 359 —

```ts
const currentFactsRows = await this.entryRepo.findRecentByEntityId(entityId, 100);
```

`doRunHeal` is the only maintenance path using `findAllByEntityId`. `recentEvents` is capped at 20 and `allTasks` is naturally small, so the two unbounded fact arrays are the entire problem.

## Observed failure

Corpus: 168 ingested documents, single entity (`business`).

| Row class | Count |
|---|---|
| `immutable_document` facts → `documentAnchors` | **2560** |
| `librarian_inferred` facts → `healCandidates` | 31 |

`documentAnchors` dominates by ~80×. Every heal pass serializes all 2560 anchors (`id`, `title`, `source_ref`) into the prompt, even though the model's job concerns only the 31 mutable candidates.

Failure 1 — output truncation at the tier's 8192-token cap:

```
Model response truncated at the 8192-token limit
  at MaintenanceService.doRunHeal (MaintenanceService.ts:491)
  at MaintenanceService.runHeal (MaintenanceService.ts:139)
```

Failure 2 — the obvious mitigation is not available. Raising the caller's max-output-tokens hits a hard model ceiling (Amazon Nova Pro):

```
The maximum tokens you requested exceeds the model limit of 10000.
```

That is the crux: the tier was already at 8192 of a hard 10000 maximum, so **at most ~22% more headroom exists**, against a prompt term that grows without bound. Config tuning cannot fix this class of bug — it only postpones it, and here it barely postpones it at all.

## Why this matters

Once a corpus crosses the threshold, `heal` fails on every run and stays failed. There is no partial progress and no self-recovery, because each attempt re-sends the same oversized prompt. Maintenance silently stops being performed while ingestion keeps enlarging the very input that broke it.

## Suggested direction

1. **Bound `documentAnchors`.** The largest and cheapest win. 2560 anchors for 31 candidates is almost all waste — consider selecting only anchors actually referenced by the candidates under review, or a bounded recent/relevant window as `runLibrarian` already does.
2. **Chunk `healCandidates`.** Process in batches with a configurable batch size (mirroring `runOntologyBackfill`'s `batchSize`), merging results across batches, so one pass cannot exceed a model's output budget.
3. **Fail loudly on the size precondition.** If a prompt is going to exceed the provider's limits, a clear up-front error naming the corpus size beats an opaque truncation or a provider 400.

Happy to send a PR if a preferred shape is settled — (1) alone would resolve the immediate breakage, but (2) is what makes it durable.

## Environment

- `@equationalapplications/core-llm-wiki` v4.22
- Model tier under test: `amazon.nova-pro-v1:0` via Bedrock Converse, hard output limit 10000 tokens
- Discovered during a full-corpus ingest validation run in a downstream consumer (`aws-cloud-agent`); ingest itself completed cleanly (168/168 documents, 0 failures), which is what first made a full-corpus `heal` reachable.

equationalapplications:
## Summary

Spec: docs/superpowers/specs/2026-08-04-maintenance-bounding-and-dependency-hygiene-design.md 

- Adds `BoundedLlmCall.runBatched`, a dependency-free batch sizer/splitter with truncation detection, sticky downward adaptation, and a skip-and-cooldown-stamp fallback for facts that can't be processed even alone.
- `doRunOntologyBackfill` and `doRunHeal` now consume `runBatched` instead of building one oversized prompt/response per pass. Heal also scopes its contradiction anchors to keyword-relevant hits (capped at 50) via new `findHealCandidatesByEntityId`/`findAnchorRowsByIds` repository queries, instead of loading every fact and every `immutable_document` row for the entity.
- `SearchService.sync()` serializes rebuilds behind a promise chain, vacuums explicitly with `autoVacuum: false`, and degrades to `console.warn` instead of an unhandled rejection — fixes the concurrent-rebuild race that crashed MiniSearch.
- Removes 4 dead `package-lock.json` files, pins `packageManager: pnpm@10.33.2`, bumps `minisearch` to `^7.2.0`, and refreshes/pins remaining dependency advisories via `pnpm.overrides` (`tar`, `postcss`, `js-yaml`, `shell-quote`).
- `LLMProvider.maxOutputTokens?` (sizing hint) and `OntologyBackfillResult.skipped` are additive-only API changes — confirmed via `packages/core/dist/index.d.ts` diff.

Closes #63, #64, #65.

## Test plan
- [x] `pnpm test` — 1052/1052 tests pass across all packages
- [x] `pnpm --recursive typecheck` — clean
- [x] Clean install from scratch (`rm -rf node_modules && pnpm install --frozen-lockfile`) verified
- [x] `pnpm --filter @equationalapplications/core-llm-wiki build` — `index.d.ts` diff confirms additive-only API surface


equationalapplications:
## /fix-pr follow-up

**Commit:** `6f11846`

### Review resolution

**Blocking / inline threads**

- **`SearchService.test.ts` — stale-clobber test proved nothing** — **Fixed.** Correct as reported: `sync()` does all its work inside a `.then()`, so nothing reads synchronously and the mutation landed before the first read. Added a `firstRead` signal to `makeScriptedRepo` and gated the mutation on it. Verified by temporarily unchaining `sync()` — the amended test goes red, the original did not.
- **`MaintenanceService.ts` — mid-pass abort discarded committed counters** — **Already fixed** in `17c2213`, before this review was posted (the review ran against `498c41e`). Thread is resolved.
- **`HANDOFF.md` — stale planning metadata** — **Fixed.** Added a "historical document" banner, dated the "current state" heading to 2026-08-04, and qualified the 17-of-36 Dependabot figure as a snapshot that only moves after merge and re-scan.

**Nitpicks**

- **`BoundedLlmCall.ts` — split sub-batches skip `maxPromptChars`** — **Fixed.** Valid and still present. Split chunks now route through `trim`. Worth noting the reasoning is stronger than "heal is not strictly safe": `buildPrompt` is not monotonic in batch size, because heal re-selects up to `HEAL_MAX_ANCHORS` anchors per batch, so a sub-batch can match anchors the parent did not and build a *longer* prompt from *fewer* items. New test constructs exactly that and asserts no multi-item batch exceeds the cap; it fails without the fix (8 over-cap sends).
- **`EntryRepository.findAnchorRowsByIds` — unchunked IN list** — **Fixed.** Chunked at `this.chunkSize`, matching `findMetadataByIds`. The 200-id ceiling was a property of the caller, not the method.
- **`packages/react/package.json` — React patch downgrade** — **Fixed, and it was worse than flagged.** Not just `packages/react` 19.2.5 → 19.2.0: the apps also went from `^19.2.6` to a hard pin at 19.2.0. Nothing required it — `react-native@0.83.6` peers on `^19.2.0`, and 19.2.8 is current stable. Aligned the workspace on **19.2.8** and added a `react`/`react-dom` override. That override also fixes a real problem the version skew was hiding: two React copies were resolving (`packages/expo` picked up 19.2.0 through react-native's peer), which produced an unmet `react-dom` peer warning. Now exactly one copy resolves, no warnings.
- **`package.json` — unbounded `>=` overrides** — **Fixed.** Changed `tar`, `postcss`, `js-yaml`, `shell-quote` to `^`. Lockfile regenerated; `--frozen-lockfile` install verified.
- **`healAnchorBounding.test.ts` — assert the requested search limit** — **Fixed.** The mock ignores its limit argument, so the tests only proved the post-retrieval cap. Now also asserts the limit passed to `searchKeyword` is finite and within `[HEAL_MAX_ANCHORS, HEAL_MAX_ANCHORS * 10]`.
- **`ontologyBackfill.test.ts` — `makeBackfillWiki` ignores `opts.batchSize`** — **Fixed by removing the param.** All three callers passed `batchSize === factCount`, and every count is under the default of 25, so forwarding it would have changed no behavior. Deleting the dead parameter is the honest fix.
- **Design doc — `runBatched` signature drift** — **Fixed.** Documents the async `buildPrompt`, `parse(responseText, batch)`, and `onSkip`.
- **`BoundedLlmCall.ts` — `trim` drops one item per iteration** — **Not applied; already fixed** in `17c2213`. `trim` binary-searches the prefix (`~log2(n)+1` builds, versus 62 for a 64→3 trim). The suggested halving would now be a regression, since it overshoots and sends smaller batches than the cap allows. The related `initialBatchSize` cap is unnecessary at binary-search cost: a 1M-token ceiling costs ~13 builds, not thousands.
- **`SearchService.sync` — coalesce queued syncs per key** — **Not applied.** Real but low-value, and the timing is wrong: this PR exists to fix a rebuild race, and adding a second concurrency mechanism to the same method in the same change is how the next race gets introduced. The burst cost is a rebuild per call, which is correct-but-redundant, not incorrect. Better as a follow-up with its own tests.
- **`MaintenanceService.runHeal` — unbounded LLM calls per pass** — **Not applied; out of scope.** Correctly identified, and I agree it is a real ceiling. Bounding it properly needs a per-pass `LIMIT` plus a checkpoint/cooldown column and a host-driven pass loop — a schema change and a new convergence contract, which is a design change rather than a review fix. Filing as a follow-up issue.

### Verification
- typecheck — **pass** (`pnpm -r typecheck`, 8 packages clean)
- lint — **N/A** (no lint script in this workspace)
- tests — **pass** (`pnpm test`) — 1059 total, all green. Core 790 (was 788; +2 new tests)
- install — **pass** (`pnpm install --frozen-lockfile`), no peer warnings
- build — **pass** (`pnpm --filter @equationalapplications/core-llm-wiki build`), public API unchanged

equationalapplications:
## /fix-pr follow-up

**Commit:** `acef3f2d646d4727607fdf4b5147816f64ddf04a`

### Review resolution
- **`packages/react/package.json` L80-81 — React override conflict (CodeRabbit)** — **Fixed.** The finding was valid and turned out to be broader than React. The root `package.json` carried a `pnpm.overrides` block alongside the `overrides:` block in `pnpm-workspace.yaml`; pnpm 10.33.2 applied **only** the package.json set, so every advisory pin in the workspace file was inert — the lockfile still resolved `esbuild@0.27.7` and `uuid@7.0.3` despite workspace overrides forbidding both, and the two files disagreed on React (19.2.0 vs 19.2.8).
  - Consolidated all overrides into `pnpm-workspace.yaml` as the single source of truth and aligned `react`/`react-dom` on `19.2.8`.
  - Removed `pnpm.overrides` from the root `package.json` (plus a comment in the workspace file warning against re-adding it).
  - Re-expressed the former package.json blanket pins (`tar`, `postcss`, `js-yaml`, `shell-quote`) as ranged pins so they cannot clobber legitimate majors — `js-yaml` correctly still resolves `3.15.0` for its 3.x consumer.
  - Regenerated `pnpm-lock.yaml`; the now-effective overrides resolve `esbuild@0.28.1` and `uuid@11.1.1`.
  - Files: `pnpm-workspace.yaml`, `package.json`, `pnpm-lock.yaml`

### Verification
- frozen install — pass (`pnpm install --frozen-lockfile`)
- typecheck — pass (`pnpm -r typecheck`, 10 projects)
- tests — pass, 1059/1059 (`pnpm test`)
- build — pass (`pnpm build`)
- audit — no known vulnerabilities (`pnpm audit --prod`)

equationalapplications:
## /fix-pr follow-up

**Commit:** `2a9177b`

### Review resolution
- **BoundedLlmCall.ts `isTruncationError` too broad (Copilot)** — **Fixed:** added an `EXCEEDS_LIMIT_PATTERN` exclusion so "requested max tokens exceeds the model/context limit" messages are no longer classified as truncation (splitting can't fix a config error — the adapter keeps requesting the same max tokens). Added regression tests covering both example message shapes. (files: `packages/core/src/services/BoundedLlmCall.ts`, `packages/core/__tests__/BoundedLlmCall.test.ts`)

All other review threads (CodeRabbit: handoff doc metadata, SearchService serialization test gap, ontology-backfill abort counter loss, React/pnpm-workspace version mismatch) were already resolved in earlier commits on this branch.

### Verification
- lint — N/A (no root lint script)
- tests — pass (`pnpm test`, 1052/1052 across all packages + integration)

PR #71, opened 2026-08-06
feat(core): export chunkText, safeSlice, and default chunking constants

### LINKED ISSUE #70: Export chunkText and safeSlice from core-llm-wiki's public API
## What

\`chunkText\` (and its helper \`safeSlice\`) are used internally by the package to split
document text into chunks before embedding, but they are not part of the public API —
\`dist/index.d.ts\` exports \`createWiki\`, \`parseEmbedding\`, \`formatContext\`, and others,
but not the chunker itself. Source: \`dist/chunk-*.mjs\`.

## Why

A downstream consumer (aws-cloud-agent's cortex judge, spec
\`2026-08-05-cortex-judge-design.md\`) needs to re-chunk a source document at read time and
rank the resulting chunks against a fact's stored embedding, to recover roughly the passage
a fact was extracted from. That only works if the re-chunking matches the chunking the
library used at ingest time.

Because \`chunkText\` isn't exported, we've had to port the algorithm verbatim into our own
codebase (\`src/judge/chunk.ts\`, copied from \`chunk-W53E3I44.mjs\` lines 996-1064). That
works today, but it's fragile: if the library's internal chunker changes in a future
version, our copy won't track it, and re-chunking will silently start disagreeing with what
the library actually did — passage selection would degrade quietly rather than failing
loudly.

## Ask

Export \`chunkText(input: string, maxChunkLength: number, overlap: number): { chunks: string[], truncated: boolean }\`
and \`safeSlice(value: string, start: number, end?: number): string\` from the package's
public entry point, so downstream consumers can call the real implementation instead of
maintaining a parallel copy.

equationalapplications:
Closes #70. Implements `docs/superpowers/specs/2026-08-06-export-chunk-text-public-api.md`.

## What

Spec: docs/superpowers/specs/2026-08-06-export-chunk-text-public-api.md

Makes the document-chunking algorithm part of the public API of `@equationalapplications/core-llm-wiki`:

- `export { chunkText, safeSlice } from './utils/pure'`
- `DEFAULT_MAX_CHUNK_LENGTH` (12000) and `DEFAULT_CHUNK_OVERLAP` (400)
- JSDoc on all four symbols
- `packages/core/__tests__/publicExports.test.ts`
- README section documenting the utilities

No behavior change and no signature change — this is a visibility change plus documentation. `packages/expo` picks the exports up for free via its existing `export *`.

## Why

A downstream consumer needs to re-chunk a source document at read time and rank the resulting chunks against a fact's stored embedding, to recover the passage a fact was extracted from. That only works if the re-chunking matches what the library did at ingest. Without a public export, the consumer had ported the algorithm verbatim, where it would silently drift from the library's behavior and degrade passage recovery with no error.

## Notes on deviations from the spec

1. **Constants live in `packages/core/src/utils/chunkingDefaults.ts`, imported by `IngestionService` and re-exported from `index.ts`** — the spec's snippet declared them in `index.ts`. Defining them once and importing at the point of use makes the anti-drift goal structural rather than conventional: the exported constant and the ingest-time default are the same binding. They initially landed directly in `IngestionService.ts`; review noted that this put the service module's runtime import graph (`PromptService` et al.) on the path of anyone importing just the constants, so they moved to a dependency-free module.

2. **Three literals replaced, not two** — the spec named two sites in `ingestDocument`, but the sanitizer fallback for a non-finite/negative `chunkOverlap` override was a third `400`. It now uses `DEFAULT_CHUNK_OVERLAP` too.

The overlap-clamp helper stays internal (spec Decision 2); the README documents it as a caveat for anyone re-chunking under a custom config. Note the clamp is evaluated on every `ingestDocument` call against the resolved overlap — it is a no-op at the shipped defaults, but bites when a custom `maxChunkLength` alone leaves the resolved overlap too large.

## Verification

- `tsc --noEmit` clean
- `vitest run` — 67 files, 816 tests passing
- `npm run build` succeeds; all four symbols present in `dist/index.d.ts` with JSDoc intact

### DIFF
--- a/docs/superpowers/specs/2026-08-06-export-chunk-text-public-api.md
+++ b/docs/superpowers/specs/2026-08-06-export-chunk-text-public-api.md
@@ -113,7 +113,10 @@ As implemented, `packages/core/README.md` also gained a "Chunking Utilities" sec
 documenting `chunkText`/`safeSlice`/the two constants for external consumers, including a
 note that `chunkText`'s per-chunk overlap is a maximum (a chunk repeats fewer than `overlap`
 characters when the previous chunk was shorter than that), and that `ingestDocument`'s
-overlap-clamp (Decision 2) still applies only to non-default configs.
+overlap-clamp (Decision 2) is evaluated on every call, including when only `maxChunkLength`
+is customized — it's a no-op for the shipped defaults, but reduces the effective overlap to
+`maxChunkLength - 1` whenever the resolved overlap (default or caller-supplied) would
+otherwise be too large.
 
 ## Minor Considerations
 
@@ -133,11 +136,12 @@ prefix change.
 
 `chunkText`, `safeSlice`, and the two new constants are moving from internal-only to public
 API surface consumed by at least one external package (aws-cloud-agent). Add JSDoc comments
-to all four at their definition sites (`packages/core/src/utils/pure.ts` for the functions,
-`packages/core/src/index.ts` for the constants) covering: what each does, parameter/return
-semantics for the functions, and — for the constants — that they are the values
-`IngestionService` uses when a caller doesn't override `maxChunkLength`/`chunkOverlap`. This
-is documentation only; no behavior or signature change.
+to all four at their definition sites (`packages/core/src/utils/pure.ts` for the functions;
+`packages/core/src/utils/chunkingDefaults.ts` for the constants — relocated there from
+`IngestionService.ts` so importing them doesn't pull in `IngestionService`'s runtime import
+graph) covering: what each does, parameter/return semantics for the functions, and — for the
+constants — that they are the values `IngestionService` uses when a caller doesn't override
+`maxChunkLength`/`chunkOverlap`. This is documentation only; no behavior or signature change.
 
 ## Non-Goals
 
--- a/docs/superpowers/specs/2026-08-06-export-chunk-text-public-api.md
+++ b/docs/superpowers/specs/2026-08-06-export-chunk-text-public-api.md
@@ -1,9 +1,10 @@
 # Spec: Export `chunkText` / `safeSlice` from the Public API
 
 **Date:** 2026-08-06
-**Status:** Draft
+**Status:** Implemented
 **Packages:** `core-llm-wiki` (primary), `expo-llm-wiki` (re-export, no code change needed)
 **Issue:** [equationalapplications/expo-llm-wiki#70](https://github.com/equationalapplications/expo-llm-wiki/issues/70)
+**PR:** [equationalapplications/expo-llm-wiki#71](https://github.com/equationalapplications/expo-llm-wiki/pull/71)
 
 ---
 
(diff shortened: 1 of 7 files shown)


equationalapplications:
## /fix-pr follow-up

**Commit:** `edce8db`

### Review resolution
- **CodeRabbit — `pure.ts` empty-input validation docs** — Fixed: qualified `chunkText`'s JSDoc (`packages/core/src/utils/pure.ts`) and `packages/core/README.md` to state the `maxChunkLength`/`overlap` constraints and "Throws if" behavior apply only to non-empty input; empty/whitespace-only input returns `{ chunks: [], truncated: false }` without validating those params.
- **CodeRabbit — spec: overlap clamp description** — Fixed: reworded `docs/superpowers/specs/2026-08-06-export-chunk-text-public-api.md` to state the clamp is evaluated on every `ingestDocument` call (including when only `maxChunkLength` is customized), and is a no-op only for the shipped defaults.
- **CodeRabbit — spec: constant doc location reference** — Fixed: updated the spec's JSDoc-location reference; since the constants were also relocated (see next item), it now points at `packages/core/src/utils/chunkingDefaults.ts`.
- **CodeRabbit / Copilot — `IngestionService.ts` constant coupling** — Fixed: moved `DEFAULT_MAX_CHUNK_LENGTH`/`DEFAULT_CHUNK_OVERLAP` into a new dependency-free `packages/core/src/utils/chunkingDefaults.ts`. `IngestionService.ts` now imports from it, and `index.ts` re-exports directly from the new module, so importing the constants from the package entry point no longer executes `IngestionService.ts` (and its `PromptService` import). The anti-drift property (constant and ingest-time default are the same binding) is preserved.
- **Copilot — `safeSlice` fractional/NaN index handling** — Not applied: this was a suppressed (non-actionable) Copilot comment, not a posted review thread. `start`/`end` are typed `number` and all current call sites pass integers; adding NaN/fractional normalization is a defensive-robustness change outside this doc/export-visibility PR's scope. Can be filed as a follow-up if a real caller needs it.

### Verification
- typecheck — pass (`pnpm --filter @equationalapplications/core-llm-wiki typecheck` → `tsc --noEmit`)
- tests — pass (`pnpm --filter @equationalapplications/core-llm-wiki test` → `vitest run`, 67 files / 816 tests)
- build — pass (`pnpm --filter @equationalapplications/core-llm-wiki build` → `tsup`; verified `DEFAULT_MAX_CHUNK_LENGTH`/`DEFAULT_CHUNK_OVERLAP` with updated JSDoc present in `dist/index.d.ts`)
- lint — N/A (no lint script in `packages/core/package.json`)

equationalapplications:
## Correction to the /fix-pr follow-up above

A self-review of `edce8db` found that my previous comment overstated what was fixed. Correcting the record.

**Commit:** `60149fb` (docs only — no source changes)

### What was wrong

The earlier summary marked the overlap-clamp wording **Fixed**. It was fixed in one of the three places that carried it. Two were missed:

1. **`packages/core/README.md:826`** still read *"when a caller overrides `chunkOverlap`, `ingestDocument()` additionally clamps it…"* — the exact claim CodeRabbit's first review thread flagged, which named three sites (the `IngestionService` JSDoc, the spec, and this README block). That thread was auto-marked *"✅ Addressed in commits e43c48e to fc130e4"*, but `git show --stat` on both commits shows they touched **only the spec file** — the README and the JSDoc were never actually updated. I trusted the resolved flag instead of verifying the named sites.

2. **Spec `Decision 2`** carried the same inaccurate claim, which left the document contradicting itself: my edit to the Testing section cited *"overlap-clamp (Decision 2) is evaluated on every call"* while Decision 2 itself said the opposite.

Additionally, the spec's **Solution** section still showed the constants declared in `IngestionService.ts` with `index.ts` importing from there — stale after the move to `utils/chunkingDefaults.ts`, which I only reflected in the *Minor Considerations* note.

### Fixed in `60149fb`

- README clamp note and spec Decision 2 both now state the clamp is evaluated on every call against the *resolved* overlap whatever its source, is a no-op at the shipped defaults, and bites when a custom `maxChunkLength` alone leaves the resolved overlap too large (`maxChunkLength: 100` + default `400` → effective `99`).
- Both now also note the clamp is only observable at ingest: passing the unclamped pair to `chunkText` throws, since it requires `overlap < maxChunkLength`. So the practical failure mode for a misled consumer is a loud error, not silent boundary drift.
- Spec Solution section updated to the shipped `utils/chunkingDefaults.ts` layout, with a note recording why the constants moved.
- Replaced brittle `file:line` references in the spec with symbol references (several had already rotted: `pure.ts:102`/`:64` → actually 141/78). The one remaining line ref, `IngestionService.ts:52-55`, is verified against HEAD.
- PR description updated — deviation note #1 still described the old `IngestionService.ts` location.

### Verification (re-run at `60149fb`)

- tests — pass (`pnpm --filter @equationalapplications/core-llm-wiki test` — 67 files / 816 tests)
- change is markdown-only; `git diff --stat` confirms no source files touched since `edce8db`

PR #75, opened 2026-08-07
fix: close heal/librarian dedupe race by moving heal's dedupe read inside its transaction (#69)

### LINKED ISSUE #69: librarian and heal may run concurrently on one entity; heal's dedupe snapshot can go stale and admit duplicate facts
## Summary

`librarian` and `heal` are allowed to run concurrently on the same entity. Both passes end by synthesizing new facts guarded by a "skip if a very similar fact already exists" check, and **heal reads its dedupe corpus outside its write transaction**. A librarian insert landing during heal's LLM call is therefore invisible to heal's dedupe check, and heal can insert a near-duplicate of it.

This is reasoned from the code, not an observed failure. Filing it so the reasoning isn't lost — it surfaced during CodeRabbit review of #68 and was deliberately left out of that PR's scope.

## The concurrency is intentional

`JobManager.acquireLock` (`packages/core/src/services/JobManager.ts:123-133`) handles `librarian`, `heal`, and `ontologyBackfill` in one branch that blocks each against its own key plus prune/reembed/import/forget — but never against each other:

```ts
case 'librarian':
case 'heal':
case 'ontologyBackfill': {
  const opKey = this.lockKeyFns[operation](entityId);
  if (this.activeMaintenanceJobs.has(opKey)) blockingOperation = operation;
  else if (this.activeMaintenanceJobs.has(this._pruneKey(entityId))) blockingOperation = 'prune';
  else if (this._isReembedActive(entityId)) blockingOperation = 'reembed';
  ...
}
```

And `packages/core/__tests__/jobs.test.ts:82` asserts it:

```ts
it('runHeal does not block runLibrarian for same entity after mutex split', async () => {
```

So this is not a missing lock. The mutex was split on purpose. The commit that would explain why predates the `@eq/wiki-core` rename (`8f1edf2`), so the original reasoning isn't easily recoverable — **that history should be established before anyone reverses the split.**

## Where the exposure actually is

The two passes are not symmetrical about when they snapshot the dedupe corpus.

Librarian re-reads it **inside** its transaction — `packages/core/src/services/MaintenanceService.ts:391` opens the transaction, `:407` reads:

```ts
const factsForDedupe = await this.entryRepo.findRecentByEntityId(entityId, 100, tx);
```

Heal reads it **before** its transaction — `MaintenanceService.ts:630`, with the transaction not opening until `:634`:

```ts
const healFactsForDedupe: Array<{ id: string; title: string }> =
  (await this.entryRepo.findInferredTitlesByEntityId(entityId))
    .filter(f => !safeDeletedSet.has(f.id));
```

Heal's window spans its entire LLM round-trip. Librarian's is a few statements. So the realistic ordering is: heal snapshots → heal calls the model (seconds) → librarian commits a new fact → heal opens its transaction and inserts a near-duplicate, because that fact was not in the snapshot.

Nothing errors. Both passes write in transactions and SQLite serializes them; the corpus just quietly accumulates redundant facts.

Note this is the same defect class as the bug fixed in #68 (commit `a2ba9ca`), where heal's dedupe snapshot went stale against heal's *own* same-pass deletions. That one was single-threaded and deterministic, so it was demonstrable. This one is a race and is not.

## Secondary: the auto-heal trigger check is stale

`packages/core/src/services/WriteService.ts:110` gates the auto-heal path on `!this.jobManager.isBlocked('librarian', entityId)`, but `maybeRunHeal` (`:145`) awaits `getCheckpoint` before reaching `tryAcquireAutoHealLock` (`JobManager.ts:257`), which only checks the heal key. A librarian pass can start in that gap.

**On its own this is harmless** and should not be "fixed" in isolation. That check is redundancy avoidance, not a safety rule — a librarian pass already calls `maybeRunHeal` when it finishes (`WriteService.runLibrarianThenMaybeHeal`, `:126`), so losing the race costs a duplicate pass, not correctness. It only matters as one more way to reach the overlap described above.

## Options

**A — Move heal's dedupe read inside its transaction.** Matches what librarian already does. Shrinks the stale window from "one LLM round-trip" to "a few statements," preserves the mutex split, doesn't change `runHeal`'s public behavior, doesn't starve heal. Cheapest real mitigation; does not fully close the race.

**B — Make librarian and heal mutually exclusive.** Structurally airtight, but reverses the deliberate split, and `wiki.runHeal()` starts throwing `WikiBusyError` while a librarian pass runs — a breaking change for hosts. Also risks starving heal: librarian fires every `autoLibrarianThreshold` (default 20) events, while heal is bounded to `HEAL_BATCH_SIZE` (25) facts per pass and needs many passes to sweep a corpus, so on a write-heavy entity heal could fall permanently behind.

**C — Leave it, document the overlap.** Defensible if the duplicates turn out to be rare or self-correcting: heal itself downgrades and deletes redundant facts on later passes, so this may already be self-limiting in practice.

Leaning A. It treats the actual exposure rather than the thing that makes the exposure reachable, and it's the only option with no API or throughput cost.

## Before acting

Worth confirming the failure is real rather than designing against a hypothesis. A targeted test — hold a librarian insert mid-heal, assert on the resulting fact count — would settle whether this is worth more than option C.

---

Context: raised by CodeRabbit on #68, declined there as out of scope. See https://github.com/equationalapplications/expo-llm-wiki/pull/68#issuecomment-5184957760 for the full review resolution.

equationalapplications:
## Summary

Spec: docs/superpowers/specs/2026-08-06-heal-librarian-dedupe-race-design.md

Closes the race where concurrent heal and librarian passes could produce near-duplicate facts. Heal read its dedupe corpus **outside** its write transaction, so a librarian insert landing during heal's LLM round-trip was invisible to heal's Jaccard check.

The fix moves heal's `findInferredTitlesByEntityId` call **inside** the transaction, after `softDeleteByIds`. With `withSerializedTransactions` serializing all transactions end-to-end, a librarian transaction either commits entirely before heal's transaction begins or entirely after it ends — there is no interleaving point for a duplicate insert.

## Changes

1. **`MaintenanceService.ts`** — Move `healFactsForDedupe` read inside the transaction after `softDeleteByIds`; drop the now-redundant `safeDeletedSet` filter (the query's `WHERE deleted_at IS NULL` handles it); pass `tx` to `findInferredTitlesByEntityId`; rewrite rationale comments.
2. **`healBounding.test.ts`** — Add a regression test that injects a librarian fact in the gap between heal's corpus read and its transaction open (red → green); update the stale comment on the delete-plus-restate test.
3. **`WriteService.ts`** — Correct the false guarantee comment: heal and librarian *can* overlap (deliberate mutex split); the check only reduces redundant passes, and correctness rests on heal's in-transaction dedupe read.

## Verification

- Regression test fails on `main` (count = 2), passes after fix (count = 1)
- All 11 heal bounding tests pass
- Full `packages/core` suite: **817 tests pass**
- `tsc --noEmit`: no errors

Closes #69

### DIFF
--- a/docs/superpowers/specs/2026-08-06-heal-librarian-dedupe-race-design.md
+++ b/docs/superpowers/specs/2026-08-06-heal-librarian-dedupe-race-design.md
@@ -1,7 +1,7 @@
 # Spec: Heal/Librarian Dedupe Race
 
 **Date:** 2026-08-06
-**Status:** Approved, not implemented
+**Status:** Implemented
 **Issue:** [#69](https://github.com/equationalapplications/expo-llm-wiki/issues/69)
 **Packages:** `@eq/wiki-core`
 
--- a/docs/superpowers/specs/2026-08-06-heal-librarian-dedupe-race-design.md
+++ b/docs/superpowers/specs/2026-08-06-heal-librarian-dedupe-race-design.md
@@ -1,0 +1,218 @@
+# Spec: Heal/Librarian Dedupe Race
+
+**Date:** 2026-08-06
+**Status:** Approved, not implemented
+**Issue:** [#69](https://github.com/equationalapplications/expo-llm-wiki/issues/69)
+**Packages:** `@eq/wiki-core`
+
+---
+
+## Problem
+
+`librarian` and `heal` may run concurrently on the same entity. Both passes end by
+synthesizing new facts behind a "skip if a very similar fact already exists" Jaccard
+check. Heal reads that dedupe corpus **outside** its write transaction, so a librarian
+insert landing during heal's LLM round-trip is invisible to heal's check, and heal can
+insert a near-duplicate of it.
+
+The two passes are not symmetrical about when they snapshot the corpus:
+
+- Librarian opens its transaction at `MaintenanceService.ts:391` and reads the corpus
+  at `:407`, passing `tx`. Its exposure is a few statements.
+- Heal reads the corpus at `MaintenanceService.ts:630-632` **without** `tx`, and does
+  not open its transaction until `:634`. Its exposure spans the entire LLM call.
+
+Nothing errors. Both passes write in transactions; the corpus quietly accumulates
+redundant facts.
+
+This is reasoned from the code, not an observed failure. It is the same defect class as
+the bug fixed in #68 (`a2ba9ca`), where heal's dedupe snapshot went stale against heal's
+*own* same-pass deletions. That one was single-threaded and demonstrable; this one is a
+race.
+
+### The concurrency is intentional
+
+`JobManager.acquireLock` (`JobManager.ts:123-133`) handles `librarian`, `heal`, and
+`ontologyBackfill` in one branch that blocks each against its own key plus
+prune/reembed/import/forget — but never against each other. `jobs.test.ts:82` asserts
+it: *"runHeal does not block runLibrarian for same entity after mutex split"*. This is a
+deliberate mutex split, not a missing lock, and this spec preserves it.
+
+---
+
+## Key finding: transactions are globally serialized
+
+`withSerializedTransactions` (`packages/core/src/db/serializedAdapter.ts:57`) wraps
+`withTransactionAsync` in a promise-chain mutex so transactions on the single shared
+connection run strictly one at a time. This was added by the
+[2026-07-09 transaction serialization spec](./2026-07-09-transaction-serialization-spec.md).
+
+This changes the conclusion issue #69 reached. The issue assumed moving heal's dedupe
+read inside its transaction would only *shrink* the stale window from "one LLM
+round-trip" to "a few statements," leaving a residual race that might justify full
+mutual exclusion.
+
+It does not. It **closes** the race. With transactions serialized end to end, a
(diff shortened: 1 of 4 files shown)


equationalapplications:
## /fix-pr follow-up

**Commit:** `805792bdb4bdd0965d16677d3385f146e904f23b`

### Review resolution
- **CodeRabbit — regression test injects librarian fact only in tx-less spy branch (Major):** **Fixed.** Rewrote the test to insert `fact_lib_1` independently before heal runs, spy on `findInferredTitlesByEntityId` to return a stale snapshot for the pre-tx read, and assert `fact_lib_1` stays live while heal creates zero matching facts. Verified the test fails without the fix (heal inserts a duplicate → count 2) and passes with it (count 1). (files: `packages/core/__tests__/healBounding.test.ts`)
- **Copilot — comments hard-code MaintenanceService line numbers (630/634):** **Fixed.** Removed the line-number references and described the sequencing instead. (files: `packages/core/__tests__/healBounding.test.ts`)

### Verification
- typecheck — pass (`pnpm typecheck` in `packages/core`)
- lint — N/A (no lint script in repo)
- tests — pass (`pnpm test` in `packages/core`: 67 files, 817 tests)

Issue #74, opened 2026-08-07
Document lifecycle gaps: no way to enumerate stored source_refs, detect content-identical documents, or plan a corpus in batch


equationalapplications:
## Context

Observed while building a Node/Lambda host on `@equationalapplications/core-llm-wiki@5.1.0`. The host walks a document inbox (S3 prefix / local dir), derives a `source_ref` from each document's path, and ingests into two entity namespaces via `ingestDocument`. On top of that it runs a contradiction detector and a human review queue over the resulting facts.

Three separate gaps showed up while fixing one live defect — the same document present at two inbox paths, ingested twice, producing duplicate facts that the detector then surfaced as contradictions of themselves. All three are facets of the same thing: **the library models a document as an opaque caller-supplied identity, and offers no lifecycle API over the set of documents an entity currently holds.**

Filing as one issue since they're related; happy to split, and happy to contribute PRs for whichever shapes you'd want.

---

## 1. No way to enumerate stored `source_ref`s, and `forget` is the only retirement primitive

`forget(entityId, { sourceRef })` is the only way to retire a document's facts. To use it, a host has to know *which* refs are stored — and nothing on `WikiMemory` exposes that. `hasChanged` only answers a question about a ref you already have in hand.

This blocks the general case of reconciling stored state against a live source: "which documents does this entity hold that no longer exist in the inbox?" is unanswerable through the public API. The only way through is raw SQL against the library's own tables from host code, which couples the host to the schema and breaks silently on upgrade.

Secondary problems with `forget` itself:

- **No dry run.** It returns `{ deleted: { entries, tasks } }` *after* deleting. A host that wants to show an operator the blast radius before acting, or gate the deletion on a threshold, can't.
- **Hard delete only.** There's no soft-retire that `heal`/`prune` would collect later, so retirement can't participate in the normal maintenance lifecycle.
- **No signal to downstream consumers.** Any host layer keyed on fact ids — review queues, annotations, judgments — is left with dangling references and no event to react to. The outbox seems like the natural place for this.

Rough shapes:

```ts
listSourceRefs(entityId: string): Promise<Array<{
  sourceRef: string;
  sourceHash: string;
  factCount: number;
  lastIngestedAt: number;
}>>;

forget(entityId, params, opts?: { dryRun?: boolean }): Promise<{ deleted: {...} }>;
```

## 2. No content identity — nothing detects two `source_ref`s carrying identical content

[...]

PR #78, opened 2026-08-07
feat(core): source-ref lifecycle, duplicate detection, batched hasChanged (#74)

equationalapplications:
## Summary

Spec: docs/superpowers/specs/2026-08-07-source-ref-lifecycle-design.md

Adds three additive public APIs to `WikiMemory` for issue #74, all backed by TDD with regression guards for the spec's load-bearing concerns:

- **§1 `listSourceRefs` + `forget({ dryRun })`**: enumerate an entity's source refs with live fact counts and latest hashes; preview a `forget` call read-only without acquiring the lock or firing outbox/embedding hooks.
- **§2 `findSourceRefsByHash` + `ingestDocument({ onDuplicateHash })`**: detect duplicate-content source_refs and gate ingestion via `ingest` (default), `skip` (return early), or `throw` (new `WikiDuplicateHashError`).
- **§3 `findLatestSourceHashes` + `hasChanged` batched overload**: 1 + N SQL queries instead of N, with per-entry `changed` and `duplicateOf` flags.

## Critical correctness work

- **All hash-by-recency queries use `ROW_NUMBER() OVER (PARTITION BY source_ref ORDER BY updated_at DESC)`**, not `MAX(source_hash)` — the spec explicitly warns against `MAX`. Regression guard: two live hashes for one ref (one at newer `updated_at`, one at lexically-larger hash); the durable test in `hasChangedBatched.test.ts` uses `importDump` to seed the anomaly.
- **Sort uses `COLLATE BINARY`** everywhere canonical ordering matters — locale-dependent ordering would re-mint identity on every deploy.
- **`forget({ dryRun })` short-circuits BEFORE `acquireLock`**, before transaction, before outbox, before embedding hooks.
- **`ingestDocument` guard runs BEFORE `acquireLock`**, before chunking, before any LLM call.
- **`hasChanged` is strictly read-only** — no locks acquired, no rows written (asserted in test).

## Latent gap closed during implementation

`JobManager.isBlocked` previously fell through to `return false` for `forget`/`import`/`ingest`/`reembed`/`global_reembed`/`global_import`, masking lock state from any caller. Extended to mirror `acquireLock` conflicts for all six operations so callers (and the new dry-run tests) can trust `isBlocked`.

## Test Plan

- [x] `npm test` in `packages/core` — 857/857 tests across 72 files
- [x] `npm run typecheck` clean
- [x] `npm run build` succeeds (ESM + CJS + DTS) in `packages/core` and `packages/expo`
- [x] New public exports reachable from `packages/expo` via `export * from '@equationalapplications/core-llm-wiki'`
- [x] Cross-method consistency test: `listSourceRefs` hash matches `findLatestSourceHash` for every ref
- [x] Multi-live-hash regression test: single-doc and batched `hasChanged` agree when two live hashes exist for the same ref
- [x] Concurrency: `hasChanged` (both overloads) acquires no lock and writes no rows

### DIFF
--- a/docs/superpowers/specs/2026-08-07-source-ref-lifecycle-design.md
+++ b/docs/superpowers/specs/2026-08-07-source-ref-lifecycle-design.md
@@ -70,7 +70,7 @@ WHERE rn = 1
 ORDER BY source_ref COLLATE BINARY
 ```
 
-`ROW_NUMBER()` (not `RANK()` or `DENSE_RANK()`) — tie-break by `ROWID` is implicit, but deterministic for our purposes — the spec never queries under a `MAX(updated_at)` tie. `COLLATE BINARY` on the outer `ORDER BY` is the explicit signal that locale dependency is rejected.
+`ROW_NUMBER()` (not `RANK()` or `DENSE_RANK()`) — tie-break by `ROWID` is implicit, but deterministic for our purposes: no spec-defined query depends on the order within a `MAX(updated_at)` tie, only on which row wins. `COLLATE BINARY` on the outer `ORDER BY` is the explicit signal that locale dependency is rejected.
 
 **Cross-method consistency:** a regression test asserts that for any `(entityId, sourceRef)`, `listSourceRefs(entityId)[i].sourceHash === (await findLatestSourceHash(entityId, sourceRef))`. This catches the exact bug `MAX(source_hash)` would introduce.
 
@@ -96,7 +96,7 @@ interface ForgetResponse {
 1. **Skip `jobManager.acquireLock('forget', entityId)`.** Dry-run is read-only.
 2. **Read outside any `withTransactionAsync`.** Dry-run counts being off-by-N during a concurrent real forget is acceptable.
 3. **Run queries** against the live state:
-   - *Standard case:* Run a `COUNT(*)` query for entries (to avoid `O(N)` memory overhead from materializing IDs) and the equivalent `TaskRepository` count for tasks. The dry-run computes whether the post-delete state would have `memory: 0, heal: 0`; if so, it reports `metadataReset: true`.
+   - *Standard case:* Run a `COUNT(*)` query for entries (the existing `softDeleteBySource` pattern materialises affected IDs to stage per-fact outbox events — see `EntryRepository.ts:357` — but dry-run stages no events, so it skips the ID materialisation and just counts) and the equivalent `TaskRepository` count for tasks. Returns `{ deleted: { entries, tasks } }` (no `metadataReset` field). The real standard-case `forget({ sourceRef })` does NOT reset the metadata checkpoint — only `clearAll: true` does (`MaintenanceService.ts:286`). `metadataReset` is therefore false for standard dry-run and real calls alike, and true only when `params.clearAll === true`. The dry-run contract is identical.
    - *`clearAll` case:* If `clearAll: true` is passed, calculate counts using `COUNT(*)` without source filters. Returns `{ deleted: { entries, tasks }, metadataReset: true }`. The real call's return type is widened to match, returning `metadataReset: true` when a checkpoint reset actually occurs, as pretending a side-effect doesn't happen is the kind of lie that bites six months later.
    - *Unknown-ref case:* Returns `{ deleted: { entries: 0, tasks: 0 } }` (no `metadataReset` field).
 4. **Return `ForgetResponse` — the same shape for real forget and dry-run.** No fact IDs in the dry-run payload.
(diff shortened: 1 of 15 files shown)


equationalapplications:
## /fix-pr follow-up

**Commit:** `e4c40b2`

### Review resolution

- **Thread 1 (spec heading)** — Fixed: canonical-selection rule is now an `## Canonical-Selection Rule {#canonical-selection-rule}` heading; the fragment link at line 148 resolves. (`docs/superpowers/specs/2026-08-07-source-ref-lifecycle-design.md`)
- **Thread 2 (deterministic tie-break)** — Fixed: spec now specifies `id ASC` as the explicit secondary key, and `EntryRepository.findLatestSourceHash`, `findLatestSourceHashes`, and `listSourceRefs` all use `ORDER BY updated_at DESC, id ASC`. New `listSourceRefs` regression test covers live rows with equal `updated_at`.
- **Thread 3 (duplicateOf canonical stored ref)** — Fixed: in both `IngestionService.ingestDocument` and `WikiMemory.hasChanged` (batched path), `duplicateOf` is now the code-unit minimum of the set of *stored different* refs; the incoming ref is excluded from the canonical-sort set. `hasChangedBatched.test.ts` mixed-entries expectations updated to match the new canonical-selection rule.
- **Thread 4 (normalization-stable fixtures)** — Fixed: duplicate-detection fixtures now use `mail-inbox-a.md` / `mail-sent-a.md` (both normalization-stable) in `ingestDuplicateHash.test.ts` (all six sites) and `hasChangedBatched.test.ts` (the `duplicateOf` test).
- **Thread 5 (chunk source-ref batch)** — Fixed: `EntryRepository.findLatestSourceHashes` now de-duplicates `sourceRefs` and processes them in chunks of `this.chunkSize - 1` (one slot reserved for `entity_id`), keeping the bind-parameter count under SQLites `SQLITE_MAX_VARIABLE_NUMBER` (default 999).
- **Thread 6 (initialize result Map with null)** — Fixed: `findLatestSourceHashes` pre-populates the result `Map` with `null` for every requested ref before overwriting entries returned by the SQL.
- **Thread 7 (source_ref IS NOT NULL)** — Fixed: `findSourceRefsByHash` query now has `AND source_ref IS NOT NULL`; legacy rows with `null source_ref` no longer leak into the duplicate guard or `WikiDuplicateHashError.canonical`. New `findSourceRefsByHash.test.ts` regression test covers a legacy null ref.
- **Thread 8 (normalize imported source_ref)** — Fixed: `ImportExportService.doImportEntity` normalizes `fact.source_ref` using the same `normalizeSourceRef` rules as `ingestDocument` and `setup()` before persisting, so subsequent duplicate checks and the default-path soft-deletion compare canonical refs consistently. New `importDump.test.ts` regression test covers a raw `mail/inbox/a.md` import that round-trips as `mailinboxa.md`.
- **Thread 9 (selector guard !== undefined)** — Fixed: `MaintenanceService.forgetDryRun` now uses `params.entryId !== undefined || params.taskId !== undefined`, matching the real-path selector detection and rejecting explicitly provided empty-string selectors.
- **Thread 10 (forgetDryRun no-selector guard)** — Already applied (commit `41a8f34`); left as-is.

### Additional nitpicks addressed
- **`MaintenanceService.ts` duplicate `clearAll` mutual-exclusion check** — Removed the unreachable duplicate from `forgetDryRun`; `forget()` runs the same check before dispatching to dry-run.
- **`WikiMemory.ts` batched `hasChanged` JSDoc** — Added JSDoc documenting that `duplicateOf` is the code-unit minimum of *stored different* refs (never the incoming ref), that `duplicateOf` uses the DB-side normalized spelling while `sourceRef` echoes the raw caller value, and that hosts should not compare the two fields directly.

### Verification
- `pnpm --filter=@equationalapplications/core-llm-wiki test` — **pass** (863/863 tests across 72 files)
- `pnpm --filter=@equationalapplications/core-llm-wiki typecheck` — **pass** (`tsc --noEmit` clean)
- `pnpm --filter=@equationalapplications/core-llm-wiki build` — **pass** (ESM + CJS + DTS)

equationalapplications:
## /fix-pr follow-up

**Commit:** `0bdb0e2`

### Review resolution
- **`packages/core/src/WikiMemory.ts` (canonical-selection re-sort in `hasChangedBatched`)** — **Fixed:** replaced `[...others].sort(...)` with `others[0]`. `EntryRepository.findSourceRefsByHash` already returns refs ordered by `source_ref COLLATE BINARY`, so re-sorting in JS (UTF-16 code-unit ordering) was wasted work and could pick a different canonical than the DB/spec for non-ASCII refs. Comment updated to explain.
- **`packages/core/src/services/IngestionService.ts` (same re-sort in duplicate-hash guard)** — **Fixed:** same `others[0]` swap. Comment updated.
- **`packages/core/src/services/ImportExportService.ts` (silent null coercion on invalid `source_ref`)** — **Fixed:** a non-null `fact.source_ref` that normalizes to `null` now throws `Error("importDump: invalid source_ref …")` before any writes — mirrors `ingestDocument`'s `Invalid sourceRef` guard rather than silently producing a legacy NULL row. Two new tests cover invalid rejection and a continued null round-trip.

### Verification
- typecheck — pass (`npm run typecheck` in `packages/core`)
- lint — N/A (no lint script in this package)
- tests — pass (`npm test` in `packages/core`: **865/865** across 72 files; affected files: importDump, ingestDuplicateHash, findSourceRefsByHash, hasChangedBatched, listSourceRefs)

Issue #79, opened 2026-08-07
`onDuplicateHash: 'skip'` guard is vulnerable to a TOCTOU race under concurrent ingestion


equationalapplications:
## Summary

The `onDuplicateHash: 'skip'` guard currently fires *before* the per-`(entityId, sourceRef)` ingest lock is acquired. Because the check occurs outside the transaction, concurrent ingest tasks for two different new files that share the exact same content hash can race, bypassing the guard and resulting in duplicate facts written to the store.

Affects: `@equationalapplications/core-llm-wiki` (the `packages/core` sub-package of this monorepo).

## Reproduction Steps

1. Two concurrent `ingestDocument` calls are initiated for different `sourceRef`s (different paths) but identical `sourceHash`es.
2. Neither `source_ref` is currently in the store.
3. Both tasks call `findSourceRefsByHash` simultaneously.
4. Both tasks see no live reference for that hash and bypass the `onDuplicateHash: 'skip'` guard.
5. Both tasks acquire their respective per-`sourceRef` locks (since the refs differ, there is no lock contention).
6. Both tasks write to the database.

**Result:** Two distinct `source_ref`s are created in the store for the exact same hash, entirely defeating the purpose of the duplicate guard.

## Current Workaround

In the host application (`aws-cloud-agent`), operators must set `documentConcurrency: 1` to serialize the execute-phase ingestion. This completely mitigates the race condition but heavily penalizes ingestion throughput — the default `INGEST_DOCUMENT_CONCURRENCY=3` was chosen to avoid Bedrock TPS throttling, and forcing it to 1 trades correctness for throughput.

## Suggested Fixes

The library needs to guarantee that the duplicate check and the write are atomic. Potential approaches:

1. **Move the guard inside the lock:** Run `findSourceRefsByHash` *after* the lock is acquired / transaction begins. The library's prior design intentionally avoided re-checking inside the lock, but the current shape is racy for new-file ingest.
2. **Introduce a per-hash lock:** Add a new lock primitive keyed by the content hash. Two tasks with the same hash would serialize through this lock; the second task would see the first's committed write and correctly trigger the guard. Lock acquisition order must be defined to avoid deadlock with the existing per-`sourceRef` lock.
3. **Atomic SQL constraints:** Use a single `INSERT ... ON CONFLICT DO NOTHING` query keyed on the hash to natively block the bypass without requiring application-level locking. Likely requires a schema change in the store.

[...]

Issue #76, opened 2026-08-07
ci: restore required Test check on main via two-phase PR release flow


equationalapplications:
### Problem

The `Protect Main` ruleset (created 2026-08-06 alongside `.github/workflows/test.yml` in #72) requires the `Test` status check on pushes to `main`. Our `release.yml` workflow uses `@semantic-release/git` to push the release commit and tags directly to `main`, and the ruleset rejects that push because no `Test` status exists for the bot's commit yet. The 5.1.1 release is currently blocked.

### Temporary mitigation

In #77, we are temporarily deleting `.github/workflows/test.yml` and dropping the `required_status_checks` rule from the ruleset so the release can publish. That trades CI gating for an unblocked pipeline and is not the desired end state.

### Goal

Restore `Test` as a required check on `main` without breaking the automated release pipeline.

### Approach: two-phase PR release flow

Replace `@semantic-release/git`'s direct push with a two-phase flow that satisfies the ruleset:

**Phase 1 - push to `main` triggers `release.yml`:**

1. Run `npx semantic-release` against the checked-out commit (still gated on `push: branches: [main]` in `release.yml`).
2. Drop `@semantic-release/git` from `.releaserc.json` so semantic-release does not commit or push. The remaining plugins (`commit-analyzer`, `release-notes-generator`, `changelog`, `npm` with `npmPublish: false`, `exec` for the monorepo version bump) still write the bumped `package.json` files and `CHANGELOG.md` to the working tree.
3. Add a step that detects the dirty working tree, creates a branch `release/v<nextRelease.version>`, commits as `github-actions[bot]`, pushes the branch, and opens a PR via `gh pr create --base main`.
4. Add `--no-ci` to `npx semantic-release` so semantic-release skips the built-in "verify CI" step and tolerates this CI environment.

**Phase 2 - PR is merged into `main`:**

1. `test.yml` runs on the release PR and reports `Test` before merge.
2. PR is merged into `main`.
3. The merge commit triggers `release.yml` again.
4. On Phase 2, `npx semantic-release` should be a no-op for version detection (the merge commit message `chore(release): vX.Y.Z` is the marker). The actual publish step needs new logic to derive `RELEASE_VERSION` from the merged PR branch name when present, falling back to `git tag --points-at HEAD` for the existing manual `workflow_dispatch` republish path.
[...]

PR #80, opened 2026-08-08
ci: Dependabot hygiene, #79 TOCTOU race fix, #76 two-phase release flow

equationalapplications:
## Summary

Implements the approved spec: [docs/superpowers/specs/2026-08-07-dependabot-concurrency-release-hygiene-design.md](docs/superpowers/specs/2026-08-07-dependabot-concurrency-release-hygiene-design.md)

- **Dependabot hygiene**: restored `Test` workflow, closed the `fast-uri` advisory (pinned to the actually-patched 3.1.5, not the spec's stale 3.1.4), added a non-cancelled `pnpm audit --audit-level=high` gate.
- **#79 TOCTOU race fix**: added a v9 partial `UNIQUE` index on `(entity_id, source_hash)` for live rows, normalized SQLite's extended `SQLITE_CONSTRAINT_UNIQUE` code across drivers, added a per-`(entityId, sourceHash)` promise-chain lock ahead of the existing sourceRef ingest lock, and translated any raced UNIQUE violation into the same duplicate-hash outcome the pre-check produces (mode-aware: skip/throw/default-ingest). Covered by a 20-iteration × 10-concurrent-caller race-cloaking test.
- **#76 two-phase release flow**: removed `@semantic-release/git`/`@semantic-release/github` so releases no longer push/tag/publish directly from `main`; replaced with a Phase 1 (open a gated `release/vX.Y.Z` PR) → Phase 2 (on squash-merge, relocate the tag, publish packages, create the GitHub Release) workflow. Updated the `Protect Main` ruleset to require the `Test` status check.

## Test plan

- [x] `pnpm run build`, `pnpm run typecheck`, `pnpm test` all pass across the workspace
- [x] `pnpm audit --json` shows zero `fast-uri` advisories (3 pre-existing js-yaml/image-size advisories remain, out of scope — noted in commit history)
- [x] `git diff --check` clean
- [x] New/updated unit tests for the v9 migration (fresh install, idempotence, soft-delete/NULL non-collision, abort path), SQLite code normalization, hash lock ordering/isolation/failure-cleanup, and UNIQUE-violation translation by mode
- [x] `release.yml` YAML validated + all embedded bash blocks pass `bash -n`; phase-detection logic traced manually for push/workflow_dispatch paths
- [x] `Protect Main` ruleset confirmed to require `test` via `gh api`
- [ ] End-to-end release smoke test (Phase 1 PR creation → auto-merge → Phase 2 publish → GitHub Release) — can only be observed on the next real conventional-commit push to `main`; not exercisable from a local session


equationalapplications:
## /fix-pr follow-up

**Commit:** `587e58b`

### Review resolution

**CodeRabbit — actionable comments (15 threads):**

- **release.yml — concurrency key + comment (line 26)** — **Fixed.** Moved concurrency to job-level on `phase1` and `phase2`, keying the group by `release-${{ github.ref }}-1` / `-2` so the two phases have separate pending slots; comment now describes the actual queuing guarantee instead of claiming non-release pushes always queue.
- **release.yml — phase detection via merged-PR list window (line 62)** — **Fixed.** Switched to `gh api "repos/${REPO}/commits/${COMMIT_SHA}/pulls"` with `COMMIT_SHA` passed through env (no shell interpolation), removing the `--limit 10` ordering pitfall.
- **release.yml — `git status --porcelain` too broad (line 145)** — **Fixed.** Restricted the check to `CHANGELOG.md package.json packages/*/package.json` (the same paths the staging step stages).
- **release.yml — MERGED state not handled + `|| true` swallows failures (line 170)** — **Fixed.** Added a `MERGED_PR` early-exit branch and replaced `|| true` with explicit `::warning::` logs that preserve intended exit behavior.
- **release.yml — `git tag` + `--follow-tags` don't actually push (line 186)** — **Fixed.** Use `git tag -a "v${RELEASE_VERSION}" -m "..."` (annotated) and push the tag ref explicitly: `git push origin "$BRANCH" "refs/tags/v${RELEASE_VERSION}"`.
- **release.yml — Phase 2 checkout uses `github.ref` not `github.sha` (line 211)** — **Fixed.** Checkout at `github.sha` for push events so the squash-merge commit (not whatever main points at when the job starts) is tagged.
- **release.yml — `gh release create` passes whole `CHANGELOG.md` (line 318)** — **Fixed.** Extract only this version's section with an `awk` block, falling back to a placeholder if the section is empty; documented the heading-format dependency in a comment.
- **design spec — MD040 missing fence languages (line 147 + 217)** — **Fixed.** Added `text` to the error-message block and `ts` to the lock-order invariant block.
- **design spec — `acquireIngestLocks` owner (lines 181, 190, 224)** — **Fixed.** Corrected all references: `acquireIngestLocks` is a `JobManager` method, not `IngestionService`. The code sample in §B4 is now also anchored to `JobManager.ts`.
- **ingestRaceTranslation.test — pre-start insert defeats skip/throw pre-check (lines 58–75)** — **Fixed.** Refactored the mock to insert the canonical row during the FIRST `findSourceRefsByHash` call for `skip`/`throw` (the pre-check), and only pre-insert for default `ingest` mode (which has no pre-check).
- **migration9.test — soft-delete + NULL fixture doesn't exercise the WHERE clause (lines 109–139)** — **Fixed.** Added a live row that shares the tombstone's hash, so the partial `deleted_at IS NULL AND source_hash IS NOT NULL` predicate is actually exercised. Updated `findSourceRefsByHash.test.ts` to use the same hash for both rows.
- **README.md — MD040 on migration v9 error fence (line 838)** — **Fixed.** Added `text` to the fence.
- **migrations.ts — fresh DBs don't enforce v9 index (line 188)** — **Not applied.** Adding `idx_entries_live_hash` to `setupDatabase()` breaks the `outbox-atomicity` and `ingest` integration tests because one `ingestDocument()` legitimately inserts multiple live facts sharing `(entity_id, source_hash)`. The v9 index as designed is incompatible with multi-fact ingests (this is a latent design flaw in the migration, not just a setupDatabase gap). Reverted to the original behavior; fresh DBs now rely on the v9 migration's abort-on-duplicate path on the v0→v9 upgrade (which is what production DBs go through). The original CodeRabbit concern is real for greenfield deployments but out of scope for this PR.
- **IngestionService — `acquireIngestLocks` after pre-check (line 90)** — **Fixed.** Moved the lock acquisition ahead of the skip/throw pre-check, still inside the `try` block. The `finally` releases both locks for every early return.
- **pnpm-workspace.yaml — fast-uri 2.x pinned to `^3.1.5` (line 24)** — **Fixed.** Pinned `>=2.0.0 <3.0.0` to exact `2.4.4` (the first patched 2.x release).

**CodeRabbit — nitpick comments (7):**

- **test.yml — `actions/cache@v4` mutable tag (line 33)** — **Fixed.** Pinned to `actions/cache@1bd1e32a3bdc45362d1e726936510720a7c30a57 # v4.2.0`.
- **test.yml — audit step blocks build/test (lines 43–50)** — **Fixed.** Moved `pnpm audit --audit-level=high` after `pnpm test`; removed the no-op `if: success() || failure()`.
- **test.yml — `permissions: {}` + persist-credentials (lines 8–12, 19–21)** — **Fixed.** Granted `contents: read` and set `persist-credentials: false` on checkout. Left `fetch-depth: 0` as-is (the migration tests rely on it being able to re-run v0→v9 setup).
- **release.yml — workflow-level write permissions (lines 14–17)** — **Fixed.** Dropped workflow-level grants; `detect` gets read-only, `phase1` / `phase2` get the scopes they actually need.
- **CONTRIBUTING.md — workflow_dispatch republish path undocumented (lines 131–140)** — **Fixed.** Added one sentence describing the recovery path (input `release_version`, reruns Phase 2 against existing tag).
- **hashLock.test — different-hash isolation not observable (lines 52–68)** — **Fixed.** Hold HASH_A with a deferred release, assert HASH_B enters while HASH_A is still held, then release A and verify both promises settle.

### Verification

- typecheck — **pass** (`pnpm run typecheck`)
- build — **pass** (`pnpm run build`)
- tests — **pass** (`pnpm test` — 887 + 51 + 81 + 15 + 6 + 73 + 10 + 36 = 1159 tests across all packages)

equationalapplications:
## /fix-pr follow-up

**Commit:** `f8dc235`

### Review resolution
- **Thread / topic:** `_hashLockKey()` collision risk in `packages/core/src/services/JobManager.ts` — **Fixed.** Switched the key encoding from `` `${entityId}${sourceHash}` `` to `` `${entityId}\0${sourceHash}` `` per the reviewer's suggestion. Neither component can contain NUL (entityIds are user-supplied identifiers; sourceHashes are hex digests), so the separator is unambiguous. Also rewrote the misleading docstring that claimed "an entityId that happens to contain the separator" when no separator was used.
  - Files: `packages/core/src/services/JobManager.ts`, `packages/core/__tests__/hashLock.test.ts`
  - Tests: added a regression test exercising the previously-colliding pair `("ab", "c"*63)` vs `("a", "b"+"c"*63)` — holds one lock, asserts the other is unblocked, asserts the raw map keys differ. Updated the two raw-key assertions in `acquireIngestLocks` tests to use the new `` `${'e1'}\0${HASH_A}` `` key.

### Verification
- tests — pass (`pnpm --filter @equationalapplications/core-llm-wiki test` → 888/888; `__tests__/hashLock.test.ts` → 8/8 including new regression test)

equationalapplications:
## /fix-pr follow-up

**Commit:** `4e9aa8d`

### Review resolution

- **Thread on `packages/core/src/db/migrations.ts` (the reviewer’s `v9 UNIQUE index` flag) — Fixed:** You were right. The reviewer observed that a UNIQUE index on `entries(entity_id, source_hash)` conflates per-fact and per-sourceRef granularity: a single `ingestDocument` writes N facts that all share `(entity_id, source_ref, source_hash)`, and the 2nd–Nth inserts hit `SQLITE_CONSTRAINT_UNIQUE` inside the normal multi-fact path. The desired invariant is *"at most one sourceRef per (entity, hash)"* — encoded structurally, not by indexing a per-fact column.
  - **What changed:** Replaced the entries-level UNIQUE with a dedicated `source_ref_index` table that records, per `(entity_id, source_hash)`, the canonical `source_ref` currently holding that hash. The partial UNIQUE index on `source_ref_index (entity_id, source_hash) WHERE deleted_at IS NULL` encodes the sourceRef-level invariant directly. The `entries` table no longer carries any UNIQUE on hash, so multi-fact ingests work unmodified.
  - **New `SourceRefIndexRepository` (packages/core/src/repositories/SourceRefIndexRepository.ts):** `upsert`, `softDeleteByEntityAndSourceRef`, `findActiveByEntityAndHash`. No outbox events — the table is an internal invariant, fully derivable from `entries`.
  - **IngestionService:** softDeletes the prior `source_ref_index` row by sourceRef at the start of the transaction, then upserts the new canonical row. The catch-and-translate fires on the `source_ref_index` UNIQUE (not the entries UNIQUE) and looks up the canonical via `source_ref_index`.
  - **MaintenanceService.forget:** soft-deletes the `source_ref_index` row alongside the entries rows so a re-ingest can take ownership.
  - **WikiMemory.findSourceRefsByHash and `hasChanged`:** use `source_ref_index` as the source of truth for "who currently holds this hash".
  - **v9 migration:** abort check now uses `COUNT(DISTINCT source_ref) > 1` (was `COUNT(*) > 1`, which always fired on normal data). The migration creates the `source_ref_index` table, its UNIQUE index, and idempotently backfills from existing live entries — one row per `(entity, hash)`, picking the sourceRef with the earliest `updated_at`. The backfill uses `INSERT OR IGNORE` with a deterministic `sri:<entity>:<hash>` ID so re-running v9 is a no-op.
  - **Spec updated:** `docs/superpowers/specs/2026-08-07-dependabot-concurrency-release-hygiene-design.md` §B1/§B2/§B5/§Tests, the cross-cutting §Public API impact, Risk #1, and the Sequencing table now reflect the source_ref_index design.
  - **Files:** `packages/core/src/db/schema.ts` (add new table), `packages/core/src/db/migrations.ts` (rewrite v9), `packages/core/src/repositories/SourceRefIndexRepository.ts` (new), `packages/core/src/services/IngestionService.ts`, `packages/core/src/services/MaintenanceService.ts`, `packages/core/src/WikiMemory.ts`; tests `migration9.test.ts`, `ingestRaceTranslation.test.ts`, `findSourceRefsByHash.test.ts`, `hasChangedBatched.test.ts`, `ingestDuplicateHash.test.ts`, `ingest.test.ts`, `services/MaintenanceService.test.ts`, `healAnchorBounding.test.ts` updated to insert into the new table where appropriate.

### Verification

- `pnpm typecheck` — pass across the workspace (core, core-llm-tools, okf, prisma-outbox, react, schema-org, expo, integration).
- `pnpm vitest run` — **889 / 889 pass** in 75 test files (7.2s). New/adapted: `migration9.test.ts` covers fresh install, idempotence, soft-delete/NULL non-collision, abort path, backfill canonical selection; `ingestRaceTranslation.test.ts` covers race-cloaking (20×10) with the new `source_ref_index` UNIQUE.
- `pnpm --filter @equationalapplications/core-llm-wiki build` — pass (CJS + DTS bundles).

The race-cloaking test (N=10 concurrent `ingestDocument` calls per iteration × 20 iterations, `onDuplicateHash: 'skip'`) now exercises the new invariant: exactly one `source_ref_index` row per iteration; the other N-1 report `duplicateOf` pointing at the winner’s ref. This test passes against the new design; under the prior `entries(entity_id, source_hash)` UNIQUE it would have failed at the 2nd fact of every multi-fact ingest.

PR #81, opened 2026-08-08
chore(release): 5.2.1

equationalapplications:
Automated release PR. Merging this via squash will trigger Phase 2 (publish + GitHub Release).

This PR was prepared manually after the automated Phase 1 run failed with `fatal: tag 'v5.2.1' already exists` — an orphan tag from a prior partial run was blocking the workflow. The orphan tag and notes ref have been deleted, and the release branch was rebuilt from `main` via the same semantic-release → chore-commit → tag → push pipeline that Phase 1 uses.

PR #82, opened 2026-08-08
fix(ci): make release.yml Phase 1 tag push idempotent

equationalapplications:
## What

Makes `release.yml` Phase 1 robust against the failure that broke the
2026-08-08 v5.2.1 release attempt.

## Why

Phase 1's bash script ended with:

```bash
git tag -a "v${RELEASE_VERSION}" -m "v${RELEASE_VERSION}"
git push origin "$BRANCH" "refs/tags/v${RELEASE_VERSION}"
```

This failed with `fatal: tag 'v5.2.1' already exists`. The cause is
that `semantic-release`'s `verifyRelease` step creates a local
lightweight tag at HEAD as part of its core flow. `@semantic-release/git`
was removed from `.releaserc.json` (per the spec in PR #80, §C2), so the
tag stays local instead of being pushed. The next `git tag -a` then
trips on the existing local tag.

Two other failure modes that `--force` also closes:
- A stale local tag from a prior workflow run that left the runner in
  inconsistent state.
- An orphan remote tag from a prior failed run (the very state the
  v5.2.1 release is in right now — a tag at `9a5dcd2` pointing at the
  spec merge, with no associated release branch).

## Fix

```bash
git tag -d "v${RELEASE_VERSION}" 2>/dev/null || true
git tag -a "v${RELEASE_VERSION}" -m "v${RELEASE_VERSION}"
git push origin "$BRANCH" "refs/tags/v${RELEASE_VERSION}" --force
```

- `git tag -d` clears any pre-existing local tag so `git tag -a` can
  replace it.
- `--force` lets the push overwrite a stale release branch or orphan
  remote tag. The release branch is ephemeral — it only ever carries
  one chore commit, so force-pushing it destroys no review history.

The detailed comment block above the change in `release.yml` records
*why* so a future reader doesn't simplify the script back into the
buggy form.

## Verification

YAML loads cleanly (`js-yaml.load`). Behavior change is bounded to
the Phase 1 tag-and-push step; nothing else in the workflow is
touched.

## Related

- v5.2.1 release PR: #81
- Spec: `docs/superpowers/specs/2026-08-07-dependabot-concurrency-release-hygiene-design.md` §C1 step 6

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -229,8 +229,9 @@ jobs:
           # (branch up, tag rejected) would silently leave Phase 1 looking green
           # while Phase 2 has nothing to relocate.
           # --force: the orphan-tag scenario we hit on 2026-08-08 (a prior
-          # verifyRelease created a remote tag at a different commit) means the
-          # push can otherwise reject. The release branch is ephemeral, so
+          # failed Phase 1 push left a remote tag at a different commit —
+          # verifyRelease's own tag is local only, per the comment above) means
+          # the push can otherwise reject. The release branch is ephemeral, so
           # force-pushing it is safe. See PR #80 spec §C1 step 6 for context.
           git push --atomic origin "$BRANCH" "refs/tags/v${RELEASE_VERSION}" --force
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -224,7 +224,15 @@ jobs:
           # force-push is safe — no review history lives on it.
           git tag -d "v${RELEASE_VERSION}" 2>/dev/null || true
           git tag -a "v${RELEASE_VERSION}" -m "v${RELEASE_VERSION}"
-          git push origin "$BRANCH" "refs/tags/v${RELEASE_VERSION}" --force
+          # --atomic: branch and tag succeed or fail together. Phase 2 depends on
+          # the tag being present after Phase 1; without atomicity a partial push
+          # (branch up, tag rejected) would silently leave Phase 1 looking green
+          # while Phase 2 has nothing to relocate.
+          # --force: the orphan-tag scenario we hit on 2026-08-08 (a prior
+          # verifyRelease created a remote tag at a different commit) means the
+          # push can otherwise reject. The release branch is ephemeral, so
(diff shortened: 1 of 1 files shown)


equationalapplications:
## /fix-pr follow-up

**Commit:** `086d7b8` (on top of `2a89e56`)

### Review resolution
- **CodeRabbit — `release.yml:227` atomic push** — **Fixed.** Added `--atomic` to the Phase 1 push so the release branch and tag succeed or fail as one transaction. Without it, a partial push could leave the remote with the branch up but the tag rejected, and Phase 2 would silently hang because `git fetch --tags` wouldn't find the tag it needs to relocate.
  - **Not applied:** CodeRabbit's narrative also asked for per-refspec force ("force only the ephemeral `$BRANCH`, reject an existing tag unless it already references the intended commit"). The `--force` on the tag is what made the v5.2.1 push work after the orphan-tag cleanup, so removing it would regress the very bug this PR exists to fix. That second piece is a separate design decision worth its own discussion (probably `--force-with-lease` on the tag with the expected SHA) — happy to follow up but didn't want to bundle it here.
  - File: `.github/workflows/release.yml`

### Verification
- typecheck — N/A (workflow YAML change)
- lint — N/A (workflow YAML change)
- YAML parse — pass (`node -e "require(js-yaml).load(...)"` loads cleanly; `jobs: detect, phase1, phase2`)

### Status
- PR #82 is open and mergeable. The thread is marked outdated by this push; resolving it now.

equationalapplications:
## /fix-pr follow-up

**Commit:** `49d8571`

### Review resolution
- **Copilot: remote-tag origin comment inaccurate** — Fixed: reworded the comment near `git push --atomic` to correctly attribute the orphan remote tag to a prior failed Phase 1 push rather than `verifyRelease` (which only creates a local tag). (`.github/workflows/release.yml`)
- **CodeRabbit: atomic push / tag protection** — Already addressed in `086d7b8` prior to this run.

### Verification
- typecheck — N/A (workflow-only YAML change, no source touched)
- lint — N/A (no linter configured for `.github/workflows`)
- YAML parse — pass (`node -e "require('js-yaml').load(...)"`)

PR #84, opened 2026-08-08
feat(core): WikiMemory public-API extensions — listEntityIds + upsertGraph

### LINKED ISSUE #85: core: no way to enumerate entity ids present in a database
## Summary

`WikiMemory` exposes no method to ask *"which `entity_id`s exist in this database?"*. Every entity-facing method takes ids as **input**; none returns the set.

Checked against `@equationalapplications/core-llm-wiki@5.2.0` (`dist/testing-CAk9oFvw.d.ts`). Of the 26 public methods, the only two that operate without an explicit id are `runReembed(entityId?)` and `exportDump(entityIds?)` — and neither reports the ids it touched. `exportDump()` technically reveals them, but it materialises every fact in the database to do it, which makes it unusable as an enumeration mechanism.

## Motivating case

In `aws-cloud-agent` we're introducing a tier registry where a namespace is either a **singleton** (`business`, `goals`) or **scoped** — a template instantiated per subject, e.g. `codebase@owner/repo`, one per reviewed repository.

Maintenance (`runLibrarian` / `runHeal` / `runOntologyBackfill` / `runPrune`) runs on a schedule with no request context. For singletons the id list is a compile-time constant, so this works today. For scoped namespaces, the host has no way to learn which instances exist, and therefore cannot sweep them.

The workarounds are both bad:

1. **Query the library's tables directly** from the host — reaching under the abstraction, and coupling to a schema and table prefix the library owns.
2. **Maintain a parallel registry of known subjects** in host-owned tables — duplicated state that can silently drift from what the library actually holds.

## Proposed API

```ts
/** Entity ids with at least one row in this database, ascending. */
listEntityIds(): Promise<string[]>;
```

A prefix filter would cover the scoped-namespace case directly, though it isn't required:

```ts
listEntityIds(options?: { prefix?: string }): Promise<string[]>;
```

Returning counts (`Array<{ entityId: string; factCount: number }>`) would additionally let a caller skip empty namespaces, but that's a nice-to-have — the id list alone unblocks us.

## Notes

- Read-only and additive; no schema change and no breaking change.
- Related capability that *does* already exist and works well for the adjacent problem: `setOntologyManifest(entityId, manifest, opts)` lets a scoped namespace be seeded at runtime rather than only via construction-time `seedManifests`. Enumeration is the remaining gap.

equationalapplications:
Implements the spec at [docs/superpowers/specs/2026-08-14-wikimemory-public-api-extensions-design.md](docs/superpowers/specs/2026-08-14-wikimemory-public-api-extensions-design.md) (now marked **Implemented**) per the implementation plan at [docs/superpowers/plans/2026-08-14-wikimemory-public-api-extensions.md](docs/superpowers/plans/2026-08-14-wikimemory-public-api-extensions.md).

Closes [#85](https://github.com/equationalapplications/expo-llm-wiki/issues/85), [#86](https://github.com/equationalapplications/expo-llm-wiki/issues/86).

## What's new

- **`WikiMemory.listEntityIds(options?: { prefix?: string })`** — read-only enumeration of `entity_id` values, including soft-deleted-only entities (closes the decommissioned-scope leak).
- **`WikiMemory.upsertGraph(entityId, params, adapter)`** — deterministic write path. Participates in the caller's transaction (C1), no-op on unchanged scope (C2), dangling edges legal (C3), strict-mode throws (C4).
- **`WikiStrictOntologyViolation`** + **`WikiSourceRefHashCollision`** error classes.
- **Internal refactor**: `IngestionService.upsertGraphCore` extracted from `ingestDocument`'s transactional tail; `validateAndNormalizeFact` / `validateInlineEdges` gain a `strict` opt-in; `EdgeRepository.softDeleteBySourceFactIds` closes the stale-edge gap.

## Commits

- `feat(core): add WikiStrictOntologyViolation and WikiSourceRefHashCollision error classes`
- `feat(core): add strict opt-in to validateAndNormalizeFact and validateInlineEdges`
- `feat(core): add EdgeRepository.softDeleteBySourceFactIds for stale-edge cleanup`
- `feat(core): widen MetadataRepository.getDistinctEntityIds to include orphaned entities`
- `refactor(core): extract IngestionService.upsertGraphCore from ingestDocument's transactional tail`
- `feat(core): add WikiMemory.upsertGraph public method`
- `feat(core): add WikiMemory.listEntityIds public method`
- `test(core): full C1-C4 contract tests for WikiMemory.upsertGraph`
- `docs(spec): mark WikiMemory public-API extensions as implemented`
- `docs(spec): drop commitIngest from public-API extensions spec`

## Verification

- Full core test suite: **927 passed** (77 files).
- `tsc --noEmit` clean.
- Regression sentinels (`ingest`, `ingestDuplicateHash`, `outbox-atomicity`, `prune-atomicity`, `ingestRaceTranslation`, `ontologyBackfill`, `healBounding`, `healAnchorBounding`, `heal-retention-boundary`): all green.

---

*Note: this branch also carries the prior CI-fix commits for `RELEASE_BOT_PAT` (release-pr-pat-anti-recursion). They're orthogonal to the WikiMemory work but bundled on this branch.*


equationalapplications:
## /fix-pr follow-up

**Commit:** `38394a2`

### Review resolution

- **Thread 1** (`.github/workflows/release.yml` lines 184–190, comment accuracy) — **Fixed.** Rewrote the explanatory comment to accurately describe GitHub's anti-recursion guard: PRs created/updated with `GITHUB_TOKEN` DO trigger `pull_request` workflows, but the `opened`, `synchronize`, and `reopened` activity types start in an "approval-required" state. `RELEASE_BOT_PAT` is retained because it bypasses that gate so the Test workflow runs autonomously.
- **Thread 2** (`.github/workflows/release.yml` lines 190–193, `[skip ci]` blocks Test workflow) — **Fixed.** Removed `[skip ci]` from the release commit message and rewrote the comment to explain why: the release branch is `release/v*`, which matches neither the Release workflow's push trigger (`branches: [main]`) nor the Test workflow's push trigger (`branches: [main]`), so the branch push itself triggers no workflow. Adding `[skip ci]` would propagate to the PR's HEAD commit and prevent the `pull_request`-triggered Test workflow from running — the very check we need to gate the squash-merge.
- **Recovery action for existing PRs** (also under Thread 2) — **Applied.** In the existing-release PR reuse path, if the PR's head commit still carries `[skip ci]` (a pre-fix artifact), push a no-op follow-up commit so the next `pull_request` event fires the Test check. Post-fix, new PRs do not have `[skip ci]` in their head commit, so this is a no-op for them.

### Verification

- typecheck — **pass** (`pnpm run typecheck` — all 10 packages clean)
- lint — **N/A** (no project-level lint config)
- tests — **pass** (`pnpm test` — 1067 tests across 87 files: 927 in `@eq/core`, 36 in integration, 73 in react, 15 in prisma-outbox, 10 in expo, 6 in schema-org)
- YAML syntax — **pass** (parsed cleanly with `node yaml`; jobs: `detect`, `phase1`, `phase2`; phase1 contains 12 steps including the new recovery block)

Only `.github/workflows/release.yml` was modified; no code or test changes were required.

equationalapplications:
## /fix-pr follow-up

**Commit:** `265a10f`

### Review resolution
- **Spec delegation target (`docs/.../2026-08-14-wikimemory-public-api-extensions-design.md:139`)** — **Fixed:** corrected `injectionService.upsertGraphCore` → `ingestionService.upsertGraphCore` (matches `WikiMemory.ts:616`).
- **listEntityIds no-lock/no-tx contract test (`packages/core/__tests__/listEntityIds.test.ts`)** — **Fixed:** added `vi.spyOn(db, "withTransactionAsync")` and `vi.spyOn(wiki.__testAccess.jobManager, "acquireLock")`; assertions now check both spies were not called during concurrent invocations.
- **EdgeRepository sourceFactIds batched DELETE (`packages/core/src/repositories/EdgeRepository.ts:87`)** — **Fixed:** chunked the DELETE on `source_id IN (…)` to 500-row batches (reusing the chunk pattern from `TaskRepository`/`OutboxRepository`), keeping `entityId` + batch IDs well under SQLite's default `SQLITE_MAX_VARIABLE_NUMBER` (32,766). Reuses the supplied `tx` executor for every batch, preserves the zero-ID fast path, returns the sum of `result.changes`.
- **IngestionService two-pass edge resolution (`packages/core/src/services/IngestionService.ts:175`)** — **Fixed:** initial `titleIndex` now excludes facts from the current `sourceRef` (so `resolveEdges` cannot target a fact that `upsertGraphCore` step (d) is about to soft-delete), and edges are deferred to a second pass against the finalized index so forward references within the new batch resolve correctly.
- **validateInlineEdges strict-mode on non-array edges (`packages/core/src/utils/ontology.ts:117`)** — **Fixed:** now throws `WikiStrictOntologyViolation` when `strict` is enabled; non-strict still returns `[]` for parity with `ingestDocument`. Two regression tests added (strict throws; non-strict still `[]`).
- **Nitpick — source-replacement coverage (`packages/core/__tests__/upsertGraphContract.test.ts`)** — **Fixed:** added a C2-adjacent contract test: write 2 facts + 1 edge under `sourceRef`/`hashA`, then write 3 facts + 2 edges under the same `sourceRef`/`hashB`. Asserts: prior facts have `deleted_at` set, prior source edge is removed, replacement edges remain, and `superseded === 3` (2 retired facts + 1 retired edge).

### Verification
- `typecheck` — pass (`pnpm -r --filter=!@equationalapplications/integration-llm-wiki typecheck` across 9 packages, all clean)
- `lint` — N/A (no lint script in this repo)
- `tests` — pass (`pnpm --filter @equationalapplications/core-llm-wiki test` → 77 files, **930 passed**, up from 927 prior — added 3 new tests covering ontology strict-mode non-array edges and source-replacement contract)

equationalapplications:
8 findings, 3 fixed:

  1. Applied — upsertGraphCore had hardcoded tags: [] and confidence: 'certain', silently
  dropping LLM-extracted fact.tags and fact.confidence from ingestDocument. Extended the nodes
  parameter with optional tags and confidence fields; ingestDocument now forwards the extracted
  values. Added regression test.
  2. Applied — Cross-sourceRef edges (sourceId pointing to a node ingested under a different
  prior sourceRef) silently failed manifest validation because sourceIdToType only included this
  call's nodes. Now distinguishes "external to call" from "in-call with null type" and passes
  external edges through (matching spec C3 dangling-target lenient treatment). Added strict-mode
  regression test.
  3. Applied — edgesWritten reported validEdges.length instead of actual addIgnoreDuplicate
  inserts, so deterministic-id collisions falsely showed as written. Now counts real inserts via
  the boolean return.

  5 documented-skipped (with rationale):

  - seedManifests fallback gap (would deviate from spec literal text — needs spec maintainers)
  - Cross-process race throwing raw SQLITE_CONSTRAINT_UNIQUE (spec §2/C2 known-limitation,
  future schema work)
  - hasConstraints OR-vs-AND design question (would break existing test, needs spec
  clarification)
  - Empty .type field in strict-mode errors (pre-existing pattern, minor diagnostic)
  - Missing runtime tx guard in softDeleteBySourceFactIds (pre-existing base-repo pattern,
  hardening-PR scope)

equationalapplications:
## /fix-pr follow-up

**Commit:** `104ffad`

### Review resolution

- **Thread 1 (release.yml lines 215–228, recovery path can't trigger Test workflow)** — **Fixed.** Configured `git config user.name`/`user.email` inside the recovery block (it runs before the post-PR-creation identity setup), and switched the push to URL-embedded-credential form using `RELEASE_BOT_PAT`: `git push https://x-access-token:${RELEASE_BOT_PAT}@github.com/${REPO}.git "$BRANCH"`. Also exposed `RELEASE_BOT_PAT` (and `REPO`) at the step env so the recovery bash can reference them — `GH_TOKEN` alone only steers `gh`, not `git push`.
- **Thread 2 (release.yml lines 215–228, recovery nested inside OPEN_PR branch)** — **Fixed.** Lifted the recovery check out of the `OPEN_PR` arm into a shared block that fires whenever `OPEN_PR` or `CLOSED_PR` is set. A legacy closed PR with `[skip ci]` is now patched before `gh pr reopen` so the reopened Test run is no longer skipped.
- **Thread 3 (spec §1/line 49, CHANGELOG claim)** — **Fixed by removing the false claim.** The CHANGELOG is auto-generated by semantic-release on the next release; I changed the sentence to point at the spec as the source of truth and to note that the entry will appear under the `core` package on next release.
- **Thread 4 (IngestionService.ts:337 / 388 / 404, direct metadataRepo bypass)** — **Fixed.** `upsertGraphCore` now resolves effective ontology state via `this.ontologyService?.getEffectiveState(entityId, tx)` (matching the pattern in `ingestDocument` at line 170) instead of `metadataRepo.getManifest` directly. This honors the `OntologyConfig.seedManifests` contract (types.ts:65-66 — seeds persisted on first access when a tx is supplied) and the `ontologyConfig.mode` fallback. Removed the now-unused `OntologyMode`/`OntologyManifest` imports. Added a regression test (`strict seedManifests throws on out-of-manifest node type even when no DB row exists yet`) that pins the contract and asserts the seed is persisted as a side effect.

### Verification

- typecheck — **pass** (`pnpm -r --filter=!@equationalapplications/integration-llm-wiki typecheck` — 10 of 11 workspace projects clean)
- lint — **N/A** (no project-level lint config)
- tests — **pass** (`pnpm --filter @equationalapplications/core-llm-wiki test` → 77 files, **933 passed**, up from 930 prior — added 1 new test for the seedManifests contract)
- YAML syntax — **pass** (parsed with `node yaml`; jobs: `detect`, `phase1`, `phase2`; phase1 contains the new recovery block)

equationalapplications:
## /fix-pr follow-up

**Commit:** `7e8fce9`

### Review resolution
- **release.yml: discard generated release files before recovery checkout** — **Fixed** in `7e8fce9`. Replaced `git checkout "$BRANCH"` with `git checkout -f -B "$BRANCH" "origin/$BRANCH"` so the working-tree semantic-release artifacts (CHANGELOG + versions) are discarded before the recovery commit. The empty no-op commit on the existing-PR path does not need the generated files.
- **upsertGraphContract.test.ts: assert the metadata row, not the seed fallback** — **Fixed** in `7e8fce9`. Replaced the `getOntologyManifest` post-assertion with a direct query to `llm_wiki_entity_manifests` and an `expect(row).toBeNull()` assertion. The comment now explains that the strict violation throws before the caller transaction commits, so the seed write performed by `OntologyService.getEffectiveState` is rolled back along with the rejected upsert — the strict-mode enforcement is proven by the `.rejects.toBeInstanceOf` assertion above.
- **IngestionService: 500-row limit applied before the sourceRef exclusion** — **Fixed** in `7e8fce9`. Added an optional `excludeSourceRef` parameter to `EntryRepository.findRecentByEntityId` so the LIMIT applies in SQL to usable rows. The IngestionService caller now passes `sourceRef` as the exclude arg and the post-query filter is gone.

### Verification
- typecheck — pass (`pnpm run typecheck` across all 10 packages)
- tests — pass (`pnpm test` in `packages/core`: 933/933 across 77 files)

equationalapplications:
## /fix-pr follow-up

**Commit:** `144cf97`

### Review resolution

- **Thread `PRRT_kwDOSP9TPs6ZakL3` — `packages/core/src/repositories/EntryRepository.ts` lines 480–481, "Keep NULL source_ref rows in the title index"** — **Fixed.** Changed the `excludeSourceRef` clause from `AND source_ref != ?` to `AND (source_ref IS NULL OR source_ref != ?)`. NULL-source_ref rows exist in production (librarian_inferred facts at `MaintenanceService.ts:518, 744`) and belong to neither the excluded sourceRef nor the replacement set; the original clause silently dropped them from the cross-sourceRef title index built at `IngestionService.ts:167`. Added a regression test (`findRecentByEntityId with excludeSourceRef keeps NULL source_ref rows in the index`) in `__tests__/repositories/EntryRepository.test.ts`.

- **Outside-diff comment — `packages/core/src/services/IngestionService.ts` lines 416–425 / 447–452, "Track ownership for externally sourced edges"** — **Not applied.** Verified against the spec (`docs/superpowers/specs/2026-08-14-wikimemory-public-api-extensions-design.md` §2 step e), the contract test `upsertGraphContract.test.ts:453`, and the recent commit `df8c236 fix(core): preserve tags/confidence + cross-sourceRef edges + honest edgesWritten`. The spec explicitly designates `deletedFactIds` (source-node based) as the supersession scope for edges — implementing the proposed sourceRef-ownership tracking would (a) invert the documented contract, (b) break the line 416 branch that the contract test at `:453` explicitly pins as required for cross-sourceRef edges (the motivating `tier_codebase` use case), and (c) require a schema change to add a `source_ref` column to the `edges` table or equivalent DB lookup on every edge insert. The behavior the reviewer flags (a cross-sourceRef edge surviving a re-parse that omits it from `params.edges`) is the specs intended semantics: edges are tied to their source nodes lifecycle, not to the parse that wrote them, exactly matching the C3 dangling-target treatment. No change made.

### Verification

- typecheck — pass (`pnpm --filter @equationalapplications/core-llm-wiki typecheck` → `tsc --noEmit`, clean exit)
- lint — N/A (no lint script defined in core or root `package.json`)
- tests — pass (`pnpm --filter @equationalapplications/core-llm-wiki test` → 77 files, **934 passed**; up from 927 thanks to upsertGraphContract and the new EntryRepository regression test)

equationalapplications:
All 4 documentation-drift findings from the review of PR #84 / branch
  fix/release-pr-pat-anti-recursion have been fixed in the spec doc to match the shipped
  implementation:

  1. edgesWritten now documented as edgesInsertedCount (successful inserts only), not
  validEdges.length.
  2. Edge validation in step (c) now documented as upsertGraphCore's own inline manifest lookup,
  not a call to validateInlineEdges.
  3. softDeleteBySourceFactIds now documented as returning a row-count number, not a list of
  edge ids.
  4. The "only ingestDocument calls these validators" claim corrected to include the two
  MaintenanceService call sites (runHeal, runOntologyBackfill), in both places it appeared.

  No code changes needed — the implementation was correct; only the spec was out of sync.

PR #93, opened 2026-08-15
feat(core): ingest parse resilience (closes #92)

### LINKED ISSUE #92: parseJsonResponse desyncs on a single unescaped quote, voiding all sibling chunks of a document
## Summary

`parseJsonResponse` (`src/*` → compiled `dist/index.js` around line 3420 in the version currently pinned by `aws-cloud-agent`, `core-llm-wiki@5.4.0`) extracts JSON from an LLM ingest response with a hand-rolled brace-matching scanner. It tracks string boundaries by toggling a flag on every `"` character:

```js
if (ch === '"') { inString = !inString; continue; }
```

This has zero tolerance for a single unescaped `"` inside the model's raw output. If the model emits one literal `"` where valid JSON requires `\"`, every subsequent `"` in the response flips string-tracking state the wrong way, and depth-counting desyncs by the time it reaches the real closing brace. `JSON.parse` on the mis-sliced substring then throws `Expected ',' or '}' after property value in JSON at position N` — a hard failure with no recovery attempt.

## Why this happens in practice

`INGEST_SYSTEM_PROMPT` instructs the model to "Extract verbatim factual content." Real documents often contain quoted example prose (e.g. a spec with `("I told you, I hate cilantro")` as a worked example). Asked to copy that verbatim into a JSON string body, the model must re-escape the embedded quotes as `\"` — and in practice a model (observed: `zai.glm-4.7-flash` via Bedrock) sometimes doesn't. This is a known category of LLM JSON-generation failure (verbatim-copy instructions competing with JSON escaping discipline), not a defect specific to any one model.

## Why the blast radius is bigger than "one bad chunk"

`ingestDocument` chunks a document and extracts each chunk independently via `withConcurrency`, then calls `parseJsonResponse` per chunk result:

```js
const chunkResults = await withConcurrency(
  chunks.map((chunk) => async () => {
    ...
    const result = parseJsonResponse(responseText);
    return { facts: ..., ontology_updates: result.ontology_updates };
  }),
  chunkConcurrency
);
```

`parseJsonResponse` throwing propagates uncaught out of the per-chunk task. Because chunk results are aggregated before any are committed, **one malformed chunk discards the facts extracted from every sibling chunk of the same document**, not just its own. In a concrete repro (35KB / 440-line spec document, `MAX_CHUNK_LENGTH=6000`), the document split into 7 chunks; testing chunks in isolation (same system prompt, without the per-call ontology-context appendix `ingestDocument` normally attaches) showed 6 of 7 parsed cleanly. In production — where `ingestDocument` additionally appends a full ontology manifest + edge-matching instructions via `ontologyService.buildPromptContext(entityId)` per chunk — the same document failed identically across 4 separate ingest runs, consistent with the longer, per-entity prompt reliably tripping the same chunk's quote-escaping mistake at `temperature: 0`.

Net effect observed downstream: a document can silently contribute **zero facts** to the graph, forever (every retry re-derives the same chunking and the same prompt, so a deterministic failure never self-heals), while other documents in the same corpus ingest normally. The host (`aws-cloud-agent`) currently has no visibility into *which* chunk failed or why beyond the raw `SyntaxError` message — it only knows the whole document failed.

## Suggested fix directions

- **Minimum:** make `parseJsonResponse` (or a wrapper) attempt a repair pass on `SyntaxError` — e.g. detect the first point of desync and try re-escaping bare quotes inside the last opened string, or fall back to a permissive JSON5-style parse — before giving up.
- **Better:** don't let one chunk's parse failure void its siblings' already-successfully-extracted facts. Chunk extraction failures should degrade to "fewer facts ingested from this document" rather than "zero facts, silently, forever."
- **Also worth considering:** tighten `INGEST_SYSTEM_PROMPT` / `ONTOLOGY_BACKFILL_SYSTEM_PROMPT` to explicitly call out JSON-escaping discipline when the instruction says to extract content "verbatim" — the two goals are in tension and the prompt doesn't currently name that tension.

## Repro sketch

Available on request — I extracted the failing document, replicated `chunkText(text, 6000, 400)`, and called `zai.glm-4.7-flash` directly via Bedrock `converse` with the same `INGEST_SYSTEM_PROMPT` for each chunk. Isolated calls (without the ontology-context appendix) parsed cleanly for all 7 chunks on that pass, which is consistent with the failure needing the longer per-entity prompt context to reliably reproduce rather than being purely random.

## Environment

- `@equationalapplications/core-llm-wiki@5.4.0`
- Host: `aws-cloud-agent`, Writer Lambda, `op: ingest`
- Model: `zai.glm-4.7-flash` via Bedrock Converse, `temperature: 0`, `MAX_TOKENS_MED: 16384`, `MAX_CHUNK_LENGTH: 6000`

equationalapplications:
## Summary

Closes #92. Implements the spec at [`docs/superpowers/specs/2026-08-15-ingest-parse-resilience-design.md`](docs/superpowers/specs/2026-08-15-ingest-parse-resilience-design.md) with post-implementation revisions.

Three layered changes for LLM-output brittleness in `IngestionService.ingestDocument`:

1. **`parseJsonResponse` tier-2 walker with arbitration.** Single container-aware pass over raw text, running two comma-parity policies at ambiguous positions (`"X"` followed by `,`): the odd-parity pass treats the quote as structural close when bare-quote count is even (legacy heuristic), the even-parity pass flips it. When **both** policies produce a parseable candidate, we arbitrate by **minimal mutation** — the candidate that escaped fewer content quotes wins (the correct interpretation escapes exactly the true bare quotes; the swallowing interpretation must also escape the structural quotes it consumes). Short-circuit skips the second pass when the first hits no comma-ambiguity. Deterministic repairs (raw C0 control chars inside strings, stray structural quotes before container close) are pass-invariant. New errors:
   - `WikiParseError extends Error` with `{ tier: 'strict'|'repair'|'all', position, slice }`.
   - `WikiIngestEmptyError extends Error` with `parseFailures, sourceRef, chunks`.
2. **Per-chunk try/catch.** Each chunk's `buildIngestPrompt` + `generateText` + `parseJsonResponse` is wrapped. `buildPromptContext` stays outside the try (DB-systemic). Failed chunks are recorded as `ChunkFailure[]` on the result; successful sibling chunks commit.
3. **Partial-commit path.** On `0 < failedChunks < chunks`, `appendPartialFacts` appends dedup-only — no supersession and no `sourceRefIndexRepo.upsert` so `hasChanged` keeps returning true and the failed chunks retry on the next run. **Retries re-extract the whole document** (deterministic chunking reproduces the same chunk set; every chunk runs through the LLM again, with dedup at write time via normalized-title collision). Total failure throws `WikiIngestEmptyError`.

`IngestDocumentResult` widens additively: `truncated, chunks` are unchanged; `ingestedChunks, failedChunks` always present; `parseFailures?` present iff any chunk failed; `duplicateOf?` present only on duplicate-hash skip.

Prompt tightening: `INGEST_SYSTEM_PROMPT` and `ONTOLOGY_BACKFILL_SYSTEM_PROMPT` now explicitly name the \" and \n escape sequences the LLM must produce.

## Post-implementation review (2026-08-16)

The first walker returned the odd-parity pass's first parse-success, so even-parity inputs like `{"title":"He said "hi"","body":"ok"}` silently swallowed the body — the exact loss mode issue #92 targets. Fixes applied (commit `eb0f753`):

- Two-pass arbitration by minimal mutation, with unambiguous-input short-circuit.
- Deterministic C0 control-char escaping (\n, \t, \r, \b, \f, \u00XX) — a literal newline otherwise kills every candidate.
- Sibling-candidate buffer reset — after a failed stack-empty emit, later balanced spans are offered standalone, not concatenated onto the already-rejected prefix.
- Stray-quote skip — a structural `"` immediately before `}`/`]` is skipped, not emitted (emitting it opens an unterminated string).
- `IngestionService.zeroChunkResult` omits `duplicateOf` when absent, restoring the `'duplicateOf' in result` contract.
- `MaintenanceService.formatSkipError` uses surrogate-safe truncation.
- `packages/react/src/useWikiIngest.ts` widened to the full `IngestDocumentResult` contract.
- 7 regression tests covering arbitration, control-char repair, and sibling-span fallthrough.

Spec revised (commits `ceeba08`, `4fd5ef2`):

- §1.2 rewritten to specify the verified arbitration design; deterministic rule documented as provably insufficient for array-prose cases.
- §4.2 corrected: retries re-extract the whole document (not just failed chunks); cost model stated explicitly as (chunks × retry runs) LLM calls.
- Open Questions: deferred findings recorded (hash-agnostic `hasChanged` after partial commit, unbounded retry accumulation, typed errors flattened on `ChunkFailure`).

## Commits

```
4fd5ef2 docs(spec): correct §4.2 retry cost model — retries re-extract whole document
ceeba08 docs(spec): revise ingest-parse-resilience §1.2 to verified arbitration design
eb0f753 fix(core): arbitrate tier-2 repair by minimal mutation; escape C0 controls
ab024d9 fix(core): address code-review findings on ingest parse resilience
86c8d3b fix(core): delegate raw String(err) coercion to safeErrorToString in parse walker
a49866a fix(core): harden safeErrorToString for hostile Error subclasses
26c4851 fix(core): address 5 cog-review items on ingest parse resilience
99c448e fix(core): guard formatSkipError against JSON.stringify returning undefined
1f6eb08 fix(core): bound skipped-item error text in MaintenanceService callbacks
0175a4a fix(core): apply high-effort code-review fixes for PR #93
7f6b921 fix(core): address cog review on PR #93 (parse resilience, retry semantics, wording)
9b60956 docs(spec): align warn-format example with implemented tag placement
```

## Verification

- `packages/core`: **1062/1062 tests pass** (88 files). `tsc --noEmit` clean.
- `packages/integration`: **43/43 tests pass** (9 files; covers repro/reparable/all-fail/retry subsets).
- `packages/react`: **73/73 tests pass**. `tsc --noEmit` clean.

## Hosts must update

Hosts that today treat `ingestDocument` throwing as the only failure signal will, after this change, see a successful return with `parseFailures[]` set when a subset of chunks failed. Inspect `result.failedChunks` and surface `result.parseFailures[]` for observability. The in-repo React host (`packages/react/src/useWikiIngest.ts`) is updated to the widened contract.


equationalapplications:
## /fix-pr follow-up

**Commit:** `7f6b921`

### Review resolution

- **thread: pure.ts — track object-key vs value string for `isClosing`** — **Fixed.** Added `stringRole` tracking on the container stack and gated the `next === ':'` close on `stringRole === 'key'`. Also reset `expectKey` on every `,` in an object (the previous code only updated it on string close, which left the walker in value-mode after a primitive — a hidden bug surfaced by the regression test).
- **thread: pure.ts — preserve failed repair candidate for `WikiParseError`** — **Fixed.** `containerAwareRepair` now returns `{ success, failed }`; `parseJsonResponse` builds `WikiParseError` with `tier: 'repair'` and the candidate slice + parse position when no candidate parses.
- **thread: IngestionService.ts — partial path stores `sourceHash` on rows, breaking `hasChanged` retry** — **Fixed.** `appendPartialFacts` now stores `source_hash: null` on partial rows; `findLatestSourceHash` returns `null` after a partial commit and `hasChanged` keeps returning `true` for same-hash retries. Added a same-hash retry unit test in `packages/core/__tests__/ingest.test.ts` and a same-hash integration test in `packages/integration/__tests__/ingestParseResilience.test.ts` (the prior retry subset used a new hash, which masked the bug).
- **thread: IngestionService.ts — `buildIngestPrompt` inside try classifies template errors as `llm` failures** — **Fixed.** Moved the prompt build outside the per-chunk try so template/override errors surface as raw throws (the same systemic-failure argument the existing comment makes for `buildPromptContext`).
- **thread: README.md — "no API change" too broad** — **Fixed.** Replaced with wording limited to `parseJsonResponse`'s signature; documented the new `WikiParseError` fields and ingest result fields; spelled out the host-facing migration in `packages/core/README.md`.
- **thread: README.md — primary ingest example still shows only `truncated` and `chunks`** — **Fixed.** Example now shows `ingestedChunks`, `failedChunks`, `parseFailures?`, and the partial-commit retry/supersede semantics inline.
- **thread: spec — table header has 3 columns but rows have 4** — **Fixed.** Spec table now uses a 4-column header (`#`, `Scenario`, `Input`, `Assertion`) so renderers don't drop the assertion cell.
- **thread: ingest.test.ts — `sourceHashFor` produces colliding hashes** — **Fixed.** Switched to `String(i).padStart(4, '0').repeat(16)` so distinct inputs produce distinct 64-char hashes.
- **thread: ingest.test.ts — ontology stub drops prototype methods** — **Fixed.** Switched to `vi.spyOn` on the real `ontologyService` instance with `mockRestore` in the teardown.

### New regression tests

- `parseJsonResponse › tier 2 › does not close a value string on a \`:\` peek-ahead` (input `{"body":"Example: "key": value"}`)
- `parseJsonResponse › repair tier surfaces failed candidate › balanced but invalid JSON throws WikiParseError with tier=repair` (input `{"facts":}`)
- `ingestDocument — partial commit › hasChanged returns true after a partial commit (so the failed chunks retry on the same hash)`
- `ingest parse resilience — integration (issue #92) › subset: same-hash retry after partial commit succeeds in picking up the failed chunks`

The test fixtures `makeBadJson()` (unit) and `bareQuoteResponse` (integration) now use `{"facts":[` instead of the earlier `"a"b"` title — the walker now legitimately repairs bare quotes inside value strings, so the earlier fixture was no longer "both tiers reject". Unclosed array bracket is the smallest input that defeats every parser tier.

### Verification

- `pnpm typecheck` — pass (all 10 packages)
- `pnpm --filter @equationalapplications/core-llm-wiki test` — **1027/1027 pass** (86 files)
- `pnpm integration-test` — **40/40 pass** (7 files; includes the new same-hash retry test)
- No lint script in this repo (no eslint config).

equationalapplications:
Code-review complete. 8 fixes applied across 4 files, all checks green (1027 core tests + 42
  integration tests, tsc --noEmit clean).

  Fixes applied

  ┌─────┬─────────────────────────────┬────────────────────────────────────────────────────────────┐
  │  #  │            File             │                            Fix                             │
  ├─────┼─────────────────────────────┼────────────────────────────────────────────────────────────┤
  │ 1   │ MaintenanceService.ts:670,  │ Extract err.message before console.warn(format, err) to    │
  │     │ 875                         │ prevent 1MB error-payload leak via util.format             │
  ├─────┼─────────────────────────────┼────────────────────────────────────────────────────────────┤
  │ 2   │ pure.ts:135                 │ scanJsonSlice now uses safeSlice (surrogate-pair safe,     │
  │     │                             │ matching chunkText)                                        │
  ├─────┼─────────────────────────────┼────────────────────────────────────────────────────────────┤
  │     │                             │ Implicit-close branch preserves peek-ahead-skipped         │
  │ 3   │ pure.ts:252                 │ whitespace so WikiParseError.slice mirrors input           │
  │     │                             │ byte-for-byte                                              │
  ├─────┼─────────────────────────────┼────────────────────────────────────────────────────────────┤
  │     │                             │ Walker returns parsed value (RepairResult.success: unknown │
  │ 4   │ pure.ts:82 + walker         │  | null) — eliminates second JSON.parse on tier-2 success  │
  │     │                             │ path                                                       │
  ├─────┼─────────────────────────────┼────────────────────────────────────────────────────────────┤
  │ 5   │ IngestionService.ts:133     │ Hoisted buildPromptContext out of per-chunk closure (1 DB  │
  │     │                             │ read per ingest instead of N)                              │
  ├─────┼─────────────────────────────┼────────────────────────────────────────────────────────────┤
  │ 6   │ IngestionService.ts:190     │ Collapsed two-pass chunkResults into one pass that builds  │
  │     │                             │ orderedChunkFacts and counts failures together             │
  ├─────┼─────────────────────────────┼────────────────────────────────────────────────────────────┤
  │ 7   │ IngestionService.ts:175     │ Folded tag ternary into the console.warn template literal  │
  ├─────┼─────────────────────────────┼────────────────────────────────────────────────────────────┤
  │     │                             │ zeroChunkResult uses duplicateOf?: string accepting        │
  │ 8   │ IngestionService.ts:27      │ undefined (no conditional spread, no                       │
  │     │                             │ exactOptionalPropertyTypes risk)                           │
  └─────┴─────────────────────────────┴────────────────────────────────────────────────────────────┘

  Skipped (out of scope, noted for follow-up)

  - LLM-error .message may embed response body — the "NEVER include raw response" comment in
  IngestionService.ts:178 is unenforceable against provider-controlled messages; right fix is a
  sanitize hook on the provider, not filtering e.message (would silently drop real provider
  diagnostics).
  - Altitude findings (walker doing double-duty, DB-systemic classification not structural, parallel
  commit paths) — require a real refactor (commitFacts(policy), preflight(), generator-style walker);
  out of scope for this branch.
  - Cumulative-buffer / bareQuoteCount — flagged as PLAUSIBLE but no clean fix without altering the
  walker algorithm.

  Diff stats: 157 insertions, 100 deletions across 4 files. Ready to inspect with git diff or commit.

equationalapplications:
## /fix-pr follow-up

**Commit:** `1f6eb08`

### Review resolution

- **Thread: `Bound the error text before logging.`** — **Fixed.** The heal (`MaintenanceService.ts` ~L670) and ontology-backfill (~L879) `onSkip` callbacks both delegate to a new module-private `formatSkipError` helper that safely stringifies `unknown` values (handles `Error`, primitives, plain objects via bounded `JSON.stringify` with fallback) and truncates output to `SKIP_ERROR_LOG_CHARS = 4096` chars. `util.format` is no longer used for the err argument. Files: `packages/core/src/services/MaintenanceService.ts`.

### Verification

- typecheck — pass (`pnpm exec tsc --noEmit` in `packages/core` and `packages/integration`)
- lint — N/A (no lint script present in either package)
- tests — pass (`packages/core`: 1027/1027 across 86 files; `packages/integration`: 42/42 across 9 files)

equationalapplications:
## /fix-pr follow-up

**Commit:** `99c448e`

### Review resolution
- **formatSkipError always returns a string (Stability & Availability, Major).** CodeRabbit flagged that `JSON.stringify(err)` can return `undefined` for `Symbol`, `function`, or objects with `toJSON: () => undefined`, causing `text.length` to throw from `onSkip` and reject the batch. **Fixed** in `packages/core/src/services/MaintenanceService.ts` by falling through to `String(err)` whenever the JSON branch yields a non-string, and retaining the `Object.prototype.toString` guard for the case where both `JSON.stringify` and `String()` throw. Exported the helper with a doc comment so callers know the export is test-only, and added a focused regression suite covering all four coercion failure modes plus the truncation marker.
- All other review threads on this PR were already resolved (table-column-count fix in `7f6b921`, source-ref-lifecycle fixes in the same commit, parser-walker object-key role tracking in `7f6b921`, repair-candidate preservation in `7f6b921`, README compatibility clarification in `7f6b921`, primary-ingest example update in `7f6b921`, and the initial skipped-item bound in `1f6eb08`).

### Verification
- lint — N/A (no lint script in `packages/core` or `packages/integration`)
- tests — pass (`pnpm --filter @equationalapplications/core-llm-wiki exec vitest run` → 87 files / 1038 tests green; new `formatSkipError.test.ts` 11/11; integration `ingestParseResilience.test.ts` 4/4 unchanged)

equationalapplications:
## /fix-pr follow-up

**Commit:** `26c4851c9b94df6e7b405c048d70b12e4da1f163`

### Review resolution
- **IngestionService embed/notify on partial path** — **Fixed.** `appendPartialFacts` now returns inserted descriptors in the same shape `runFullUpsertGraph` returns; the post-commit `embedFact` loop runs on either path. `notifyEmbeddingPersisted(entityId, factId, null)` still stays gated on the full path (no supersession, nothing to retire). (files: `packages/core/src/services/IngestionService.ts`)
- **Non-throwing error coercion in per-chunk catch** — **Fixed.** New `safeErrorToString` helper wraps `e.message` / `String(e)` / `Object.prototype.toString.call(e)` in nested try/catch so a provider rejection whose `toString()` throws can no longer escape the catch and unwind `withConcurrency`. (files: `packages/core/src/services/IngestionService.ts`, `packages/core/src/utils/pure.ts`)
- **Provider `Error.message` no longer leaked via `console.warn`** — **Fixed.** Warn line is now a narrow `chunk N/M source [tier=X][ position=Y] (sourceRef=...)` tag. Full message stays in `parseFailures` for callers that opt in; the warn line is intentionally narrow on both branches. (files: `packages/core/src/services/IngestionService.ts`)
- **Integration test now exercises a real bare-quote case end-to-end** — **Fixed.** `bareQuoteResponse` now contains a literal `"hi"` in the body; new test `subset: 4 ok + 1 reparable bare-quote` asserts all five facts reach storage, `parseFailures` is empty, and the repaired fact preserves the `"hi"` prose. The unfixable `{"facts":[` fixture is renamed to `truncatedPayload` and the three failure-case tests now use it. (files: `packages/integration/__tests__/ingestParseResilience.test.ts`)
- **Comma-ambiguity fix (parity alone silently corrupts single bare quote before a later property)** — **Fixed.** `containerAwareRepair` now takes a `closeOnEvenParity` parameter; `parseJsonResponse` runs the walker twice under opposing policies and accepts the first parse-success. Closed `bareQuoteCount === 0` short-circuit so the parity heuristic never disagrees on a structurally unambiguous comma. Regression tests: `parsers single bare quote before a later property (comma-ambiguity)` and the new integration test above. (files: `packages/core/src/utils/pure.ts`, `packages/core/__tests__/parseJsonResponse.test.ts`)

### Verification
- typecheck — pass (`pnpm exec tsc --noEmit` clean on `packages/core` and `packages/integration`)
- lint — N/A (no lint script in monorepo root or these packages)
- tests — pass (`packages/core`: 1046/1046 tests across 88 files; `packages/integration`: 43/43 tests across 9 files)

equationalapplications:
Resolution summary

  Root cause of the tension: both fixes were patching formatSkipError directly, but the helper they
  both rely on — safeErrorToString — had the same hostile-Error hole (Angle A's finding #2). Fixing
  only formatSkipError would leave IngestionService.ts:181 (which calls safeErrorToString for the same
  reason) with the same bug. The committed 99c448e was strictly weaker than the working-tree version
  (still throws on hostile Proxy), and both were vulnerable to Error with bad message.
  
  Changes

  1. packages/core/src/utils/pure.ts — hardened safeErrorToString:
  - Added a readErrorField(e, 'message'|'name') helper that wraps property access in try/catch.
  - Error branch now: tries e.message; if not a non-empty string, falls back to e.name; if neither,
  returns '[Error]'.
  - Closes Angle A finding #2 at the helper level. Both formatSkipError and IngestionService inherit
  the hardening.

  2. packages/core/src/services/MaintenanceService.ts — tightened formatSkipError:
  - Kept the working-tree's hybrid shape (JSON for objects → safeErrorToString fallback).
  - Dropped the redundant || err === null clause — typeof null === 'object', so JSON.stringify(null) 
  === 'null' handles it cleanly in the else branch.
  - Improved the inline comment to document the never-throws contract and the JSON-vs-safeErrorToString
  decision boundary.
  
  3. Tests — added 5 cases to safeErrorToString.test.ts and 3 to formatSkipError.test.ts covering:
  Error with undefined/non-string/throwing-getter message, and Error with empty message + custom name
  fallback.
  
  Verification

  ┌─────────────────────────────────────────────────┬──────────────────────────────────────────────┐
  │                      Suite                      │                    Result                    │
  ├─────────────────────────────────────────────────┼──────────────────────────────────────────────┤
  │ safeErrorToString.test.ts (9 cases)             │ ✅                                           │
  ├─────────────────────────────────────────────────┼──────────────────────────────────────────────┤
  │ formatSkipError.test.ts (15 cases — 11          │ ✅                                           │
  │ committed + 1 Proxy + 3 Error-with-bad-message) │                                              │
  ├─────────────────────────────────────────────────┼──────────────────────────────────────────────┤
  │                                                 │ 1241/1241 pass (was 1047; +5 helper tests,   │
  │ Full core vitest                                │ +6 formatSkipError tests, + tests from prior │
  │                                                 │  commits)                                    │
  ├─────────────────────────────────────────────────┼──────────────────────────────────────────────┤
  │ Integration (pnpm test)                         │ 41/41 pass                                   │
  ├─────────────────────────────────────────────────┼──────────────────────────────────────────────┤
  │ Expo                                            │ 10/10 pass                                   │
  ├─────────────────────────────────────────────────┼──────────────────────────────────────────────┤
  │ React                                           │ 73/73 pass                                   │
  └─────────────────────────────────────────────────┴──────────────────────────────────────────────┘

  Hostile-case probe (post-fix)

  hostile proxy                    → "[unstringifiable error]"
  Error with undefined message     → "Error"
  Error with non-string message    → "Error"
  Error with throwing message getter → "Error"
  Symbol                           → "Symbol(skip)"
  function                         → "function skipped() { return 1; }"
  cyclic object                    → "[object Object]"

  Every input that previously threw now returns a bounded string.

  Working-tree state

  modified:   packages/core/__tests__/formatSkipError.test.ts   (+54)
  modified:   packages/core/__tests__/safeErrorToString.test.ts (+43)
  modified:   packages/core/src/services/MaintenanceService.ts  (rewritten)
  modified:   packages/core/src/utils/pure.ts                   (+25/-3)

equationalapplications:
All three findings verified, fixed, and gates passed:

  Findings re-verified against the codebase before applying:

  ┌─────┬────────────────────────────────────────┬─────────────────────────────────────────────────┐
  │  #  │                Finding                 │                    Verified                     │
  ├─────┼────────────────────────────────────────┼─────────────────────────────────────────────────┤
  │     │ Spec §2 warn-format example omits      │ ✓ IngestionService.ts:183-194 comment block     │
  │ 1   │ <message> and reorders (tier=repair) — │ states the message is omitted "in the interest  │
  │     │  implementation deliberately omits the │ of a single, narrow warn format"                │
  │     │  message                               │                                                 │
  ├─────┼────────────────────────────────────────┼─────────────────────────────────────────────────┤
  │ 2   │ Dangling ; when both tierTag and       │ ✓ Re-read code at lines 196-200; output would   │
  │     │ positionTag are ''                     │ have been (sourceRef=doc://spec;)               │
  ├─────┼────────────────────────────────────────┼─────────────────────────────────────────────────┤
  │     │ findByIds(liveIds, [entityId], tx) at  │ ✓ EntryRepository.ts:300-323 (findIdsBySource)  │
  │ 3   │ line 699 adds redundant AND entity_id  │ hard-codes WHERE entity_id = ? with entityId as │
  │     │ IN (?)                                 │  the first arg                                  │
  └─────┴────────────────────────────────────────┴─────────────────────────────────────────────────┘

  Fixes applied:

  - packages/core/src/services/IngestionService.ts:195-202 — Replaced tierTag/positionTag pair with a
  tags array joined by space; emits ;  only when non-empty. New format: (sourceRef=doc://spec) →
  (sourceRef=doc://spec; tier=repair) → (sourceRef=doc://spec; tier=repair position=42).
  - packages/core/src/services/IngestionService.ts:694-702 — findByIds(liveIds, undefined, tx) (was
  [entityId]); updated comment to record why the entity filter is unnecessary.
  - packages/core/__tests__/ingest.test.ts:211-226 — Test stub heuristic was hard-coded to require
  entity_id IN filtering on the findByIds branch; updated to detect the clause's presence via
  normalized.includes('entity_id IN (') so it handles both call shapes (the existing WriteService call
  sites with [entityId] and the new partial-path call without).
  - docs/superpowers/specs/2026-08-15-ingest-parse-resilience-design.md:148-158 — Replaced the two
  stale warn examples with the actual format and added a paragraph explaining the ;-emission rule and
  the deliberate message omission.

  Verification (fresh, this turn):

  core:        Test Files  88 passed (88)   Tests  1055 passed (1055)
  core typecheck: tsc --noEmit  →  exit 0 (no output)
  integration: Test Files  9 passed (9)     Tests  43 passed (43)

  Note: counts are higher than the PR description (1024/41 → 1055/43; 86/9 files → 88/9) — additional
  tests landed on this branch since the description was written; all current tests pass.

PR #97, opened 2026-08-17
fix(core): harden eight Tier-A instanceof Error checks against hostile Proxy getPrototypeOf traps (closes #96)

### LINKED ISSUE #96: formatSkipError: guard instanceof Error checks against hostile Proxy
## Context

Flagged by CodeRabbit during review of #95 (docs-only PR), on code that lives on `main` after #93 squash-merged: [review comment](https://github.com/equationalapplications/expo-llm-wiki/pull/95#discussion_r3793032531).

## Problem

`formatSkipError` in `packages/core/src/services/MaintenanceService.ts` (currently around line 117) does:

```ts
if (err instanceof Error || (typeof err !== 'object' && typeof err !== 'function')) {
  base = safeErrorToString(err);
} else {
  try {
    const json = JSON.stringify(err);
    base = typeof json === 'string' ? json : safeErrorToString(err);
  } catch {
    base = safeErrorToString(err);
  }
}
```

`err instanceof Error` performs a `getPrototypeOf` walk. If `err` is a `Proxy` whose `getPrototypeOf` trap throws, this check throws *before* any of the safe fallback paths (including `safeErrorToString`) get a chance to run — defeating the whole point of the helper, which exists specifically to never throw regardless of what hostile value reaches it via `onSkip`.

Reproduces with:

```js
const proxyErr = new Proxy({}, {
  getPrototypeOf() {
    throw new Error('proxy rejects prototype access');
  },
});
proxyErr instanceof Error; // throws
```

## Fix

Wrap the `instanceof Error` check in `try/catch`; on throw, treat `err` as non-Error and fall through to the existing `JSON.stringify` / `safeErrorToString` path. Add a regression test in `packages/core/__tests__/formatSkipError.test.ts` covering a proxy with a throwing `getPrototypeOf` trap, asserting `formatSkipError` does not throw and returns the non-Error fallback.

## Scope

Core package follow-up — out of scope for #95 (docs-only), since the file left that PR on rebase after #93 squash-merged.

equationalapplications:
Follow-up to PR #95 (docs-only after #93 squash-merged). Audit found the bug class is broader than the function CodeRabbit flagged: `err instanceof Error` invokes the getPrototypeOf trap on err, and a Proxy whose trap rejects turns the type-check into a throw point, escaping every catch around the function body. Eight sites in `packages/core/src` take `unknown` from external/plugin sources and have documented non-throwing contracts that the operator now honors.

## Sites fixed

Same idiom (single boolean at function top via `let + try/catch`) across all eight — no new exports, no helper module:

- **safeErrorToString** (site 1) — packages/core/src/utils/pure.ts
- **sanitizeRankerError** (sites 2, 3, 4 collapsed to one boolean) — packages/core/src/utils/pure.ts
- **formatSkipError** (site 5) — packages/core/src/services/MaintenanceService.ts
- **isTruncationError** (site 6) — packages/core/src/services/BoundedLlmCall.ts
- **RetrievalService inline rankerErr catch** (site 7) — packages/core/src/services/RetrievalService.ts
- **RetrievalService inline phase-2 catch** (site 8) — packages/core/src/services/RetrievalService.ts

## Test coverage

- **One new dedicated regression file** (`instanceofErrorProxyGuard.test.ts`) — single grep target for future audits of this vulnerability class.
- **Four per-file smoke tests** — appended to the existing `formatSkipError`, `safeErrorToString`, `BoundedLlmCall`, and `vectorRanker` test files, so each function's regression is discoverable when reading the function in isolation.
- **Site-8 regression test** — the original `vectorRanker.test.ts` integration covered only site 7 (rankerErr catch); a follow-up test exercises site 8 (RetrievalService phase-2 catch) directly by throwing a hostile Proxy from the embed function with no `vectorRanker` configured. See `## Follow-up changes from high-effort review` below.

## Spec & prior PRs

- Spec: [`docs/superpowers/specs/2026-08-17-instanceof-error-proxy-guard-design.md`](./docs/superpowers/specs/2026-08-17-instanceof-error-proxy-guard-design.md) (Approved → Implemented). Draft committed at f87098e; Approved revision + 5 review-driven tweaks at 146976e; implementation corrections + Implemented status at 6fc7282; review-driven spec revisions (line numbers, audit grep) at 4785cd2.
- Plan: docs/superpowers/plans/2026-08-17-instanceof-error-proxy-guard.md (gitignored; spec/code split convention)
- Originates from #96, follow-up to docs PR #95 after #93 squash-merged onto `main`.

## Verification

- `pnpm --filter @equationalapplications/core-llm-wiki test` → **1079/1079 PASS** (was 1078; +1 is the site-8 regression test)
- `pnpm --filter @equationalapplications/core-llm-wiki exec vitest run __tests__/instanceofErrorProxyGuard.test.ts` → 12/12 PASS
- Audit grep (awk-based; see spec §Verification) → `Protected: 6 / 6` — every Tier-A `instanceof Error` code site wrapped in `try/catch`, both single-line and multi-line forms

## Implementation corrections from per-task review

Per-task review caught one empirically-wrong claim in the plan's verbatim test text (commit 6a44d7c): `String(hostileProxy)` and `Object.prototype.toString.call(hostileProxy)` do NOT invoke `getPrototypeOf` and return `'[object Object]'` (not throw); `JSON.stringify(hostileProxy)` returns `'{}'`. The dedicated regression test was updated to assert the actual observable return values. Spec §Test Strategy and §Per-function smoke-tests table were also corrected.

## Follow-up changes from high-effort review

After the initial PR was opened, `/code-review high --fix` surfaced 4 additional findings. All 4 are addressed in two follow-up commits:

### Test fixes (commit 1c78bc1)

1. **site-8 regression test added** to `packages/core/__tests__/vectorRanker.test.ts` — the original integration test only exercised site 7 (rankerErr catch), so site 8 had no test that would catch a future revert. The new test throws a hostile Proxy from the embed function with no `vectorRanker` configured, so the throw goes to the outer try's catch (site 8) instead of the inner rankerErr catch (site 7). Confirmed failing before the site-8 fix (`Error: proxy rejects prototype access` at RetrievalService.ts:506) and passing after.
2. **Misleading test name corrected** in `packages/core/__tests__/instanceofErrorProxyGuard.test.ts` — renamed from `'returns the input unchanged…'` to `'wraps the input in a synthetic Error when sanitizeRankerErrors=false…'`. The function actually wraps a non-Error in `new Error(String(err))`, not return-as-is. The body comment already documented the wrap; the name was the lie.

### Spec revisions (commit 4785cd2)

3. **Stale Tier-A site line numbers** in the §Scope table — the table used pre-patch line numbers (e.g. `RetrievalService.ts:489`, `utils/pure.ts:538`) that drifted 5–17 lines because the implementation added doc comments above each guarded site. Updated to current locations: `utils/pure.ts:545`, `utils/pure.ts:591` (sites 2–4 collapsed to one boolean), `MaintenanceService.ts:125`, `BoundedLlmCall.ts:70`, `RetrievalService.ts:350`, `RetrievalService.ts:506`. Also updated the inline reference to `readErrorField` at `utils/pure.ts:576–582`.
4. **Audit grep hardened for single-line try/catch** — the previous `grep -B1` only matched the multi-line `try {` ... `instanceof Error` form; a single-line `try { x instanceof Error } catch {}` would not be counted. Replaced with an awk script that tracks whether a `try {` has been seen on any recent line and counts both forms. Also switched the final assertion from string-compare (`=`) to numeric-compare (`-eq`) so `wc -l`'s leading whitespace doesn't break the test. Verified by running the awk against synthetic input with all three forms (single-line, multi-line, orphan `try {`); only the truthy cases were counted.

## Tier-B (out of scope)

Per spec §Scope: `instanceof WikiBusyError`, `WikiParseError`, `WikiTransactionError`, `Uint8Array`, `Float32Array` checks (8 sites) excluded — values come from trusted internal sources and have no Proxy attack surface in practice.

## Also in this PR

- **README cleanup**: removed the stale 5.4.1 "Recent changes" section (belongs in CHANGELOG.md) and promoted GraphRAG to its own Key Principles bullet, separate from the seeded-ontology bullet.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -56,15 +56,33 @@ jobs:
 
       - name: Verify README GraphRAG SQL is in sync with EdgeRepository
         run: |
-          # Check WITH RECURSIVE in both EdgeRepository.ts and README.md
-          grep -q "WITH RECURSIVE" packages/core/src/repositories/EdgeRepository.ts || { echo "Missing WITH RECURSIVE in EdgeRepository"; exit 1; }
-          grep -q "WITH RECURSIVE" README.md || { echo "Missing WITH RECURSIVE in README.md"; exit 1; }
-          # Check cycle guard in both EdgeRepository.ts and README.md
-          grep -q "instr(w.visited" packages/core/src/repositories/EdgeRepository.ts || { echo "Missing cycle guard in EdgeRepository"; exit 1; }
-          grep -q "instr(w.visited" README.md || { echo "Missing cycle guard in README.md"; exit 1; }
-          # Check confidence CASE in both EdgeRepository.ts and README.md
-          grep -q "WHEN 'certain' THEN 2" packages/core/src/repositories/EdgeRepository.ts || { echo "Missing confidence CASE in EdgeRepository"; exit 1; }
-          grep -q "WHEN 'certain' THEN 2" README.md || { echo "Missing confidence CASE in README.md"; exit 1; }
+          REPO=packages/core/src/repositories/EdgeRepository.ts
+          DOC=README.md
+          # Presence checks: both files must still contain each SQL fragment.
+          for pat in "WITH RECURSIVE" "instr(w.visited" "WHEN 'certain' THEN 2"; do
+            grep -q -- "$pat" "$REPO" || { echo "Missing '$pat' in EdgeRepository"; exit 1; }
+            grep -q -- "$pat" "$DOC" || { echo "Missing '$pat' in README.md"; exit 1; }
+          done
+          # Content checks: the depth guard and final ranking clause must match
+          # semantically, not just be present — this is what catches an operator
+          # change (e.g. `<` -> `<=`) or a reordered ORDER BY going undetected.
+          extract() {
+            # $1 = file, $2 = grep pattern; strips SQL comments/template-literal
+            # prefix interpolation and collapses whitespace so the two files'
+            # differing styles (raw TS vs abridged, commented README) compare equal.
+            grep -- "$2" "$1" | head -1 \
+              | sed -E 's/--.*$//; s/\$\{this\.prefix\}/llm_wiki_/g; s/[[:space:]]+/ /g; s/^ //; s/ $//'
+          }
+          repo_depth=$(extract "$REPO" "WHERE w.depth")
+          doc_depth=$(extract "$DOC" "WHERE w.depth")
+          if [ "$repo_depth" != "$doc_depth" ]; then
+            echo "Depth guard clause differs:"; echo "  EdgeRepository: $repo_depth"; echo "  README:         $doc_depth"; exit 1
+          fi
+          repo_order=$(extract "$REPO" "ORDER BY depth ASC")
+          doc_order=$(extract "$DOC" "ORDER BY depth ASC")
+          if [ "$repo_order" != "$doc_order" ]; then
+            echo "ORDER BY clause differs:"; echo "  EdgeRepository: $repo_order"; echo "  README:         $doc_order"; exit 1
+          fi
 
       # Audit runs LAST so a single high-severity advisory introduced by
       # this PR cannot mask build / typecheck / test failures.
--- a/README.md
+++ b/README.md
@@ -24,49 +24,17 @@ Supports [Open Knowledge Format (OKF) v0.1](https://github.com/GoogleCloudPlatfo
 - **Universal Support:** Expo • React • Vite • Vue • Svelte • Node.js
 - **Core Engine:** Pure TypeScript logic with platform-specific adapters.
 
-## Recent changes
-
-### 5.4.1 — Ingest parse resilience (issue #92)
-
-- `parseJsonResponse` now tolerates bare `"` characters in LLM output via a
-  container-aware repair pass. The function's signature is unchanged
-  (`parseJsonResponse<T>(text: string): T`); the runtime contract widens —
-  the new `WikiParseError` type carries `{ tier, position, slice }` instead
-  of an opaque `Error`, and `tier: 'repair'` is now reachable for balanced
-  invalid payloads (e.g. `{"facts":}`) where the walker found a candidate
-  span that `JSON.parse` ultimately rejected.
-- `IngestionService.ingestDocument` no longer rejects the whole call when one
-  chunk fails. Sibling chunks commit; the result's `parseFailures[]` records
-  per-chunk failures. A new typed error `WikiIngestEmptyError` is thrown when
-  every chunk failed. New typed error `WikiParseError` carries `{tier,
-  position, slice}`. New result fields: `ingestedChunks`, `failedChunks`,
-  `parseFailures?`.
-- Partial-commit semantics: when some chunks fail, the document's
-  `(entity, sourceHash) → sourceRef` ownership is **not** recorded and the
-  partial rows are stored with `source_hash = NULL`. Subsequent runs with
-  the same hash see `hasChanged` return `true` and the failed chunks retry
-  on the next pass; the retry's `appendPartialFacts` dedupes against the
-  prior partial's surviving rows so titles are never duplicated. On full
-  success the next run supersedes everything atomically.
-- `INGEST_SYSTEM_PROMPT` and `ONTOLOGY_BACKFILL_SYSTEM_PROMPT` were tightened
-  to call out JSON-escape discipline explicitly.
-- **Hosts must update**: a host that today treats `ingestDocument` throwing as
-  the only failure signal will, after this change, see a successful return
-  with `parseFailures[]` set when a subset of chunks failed. Hosts should
-  inspect `result.failedChunks` and surface `result.parseFailures[]` for
-  observability. `aws-cloud-agent`'s Writer Lambda was updated in this
-  release.
-
 ## Key Principles
 
 - **Bring Your Own Inference (BYOI):** Provide one `generateText` function. The package owns prompt construction, JSON parsing, and database writes.
 - **Namespace Safe:** All tables are prefixed (default: `llm_wiki_`) — no collisions with your existing database.
 - **Multi-Entity:** Multiple independent "brains" in one database. `read()` accepts a single `entityId` string or an array for cross-namespace retrieval with per-entity `tierWeights`.
 - **Semantic Retrieval:** Supply an optional `embed()` function on `LLMProvider` to rank facts by vector cosine similarity. Falls back to MiniSearch keyword search when `embed` is absent or offline.
+- **GraphRAG:** SQL-only graph retrieval without the need for external databases such as neo4j.
+- **Seeded ontology:** Optional per-entity taxonomies (Strict, Emergent, or Off) guide librarian and ingest passes to classify facts with `okf_type` and persist structured graph edges alongside semantic and episodic memory.
 - **Offline First:** The MiniSearch fallback runs entirely in-process with no network required. The cosine similarity path requires `embed()` to vectorise the query (typically a cloud API call) but falls back to MiniSearch automatically when offline or when `embed` throws.
 - **Full Unicode Support:** UTF-8 and UTF-16 (including surrogate pairs for emoji) are fully supported. Chunks are split safely at sentence boundaries; surrogate pairs are never fragmented.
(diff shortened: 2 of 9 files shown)


equationalapplications:
## /fix-pr follow-up

**Commit:** `8f37444`

### Resolution

The failing `Test` workflow run (`pnpm run build` → `dts` step) was a pure TypeScript narrowing issue. The spec-prescribed `let isErrorLike; try { isErrorLike = e instanceof Error; } catch {}` idiom is runtime-correct, but TypeScript can't carry the narrowing across the try/catch boundary into the `if (isErrorLike)` branches — the `dts` build then errors with `Argument of type 'unknown' is not assignable to parameter of type 'Error'`. Purely a typecheck fix; runtime behavior is unchanged.

Three minimal edits (no new exports, no helper module, same inline idiom):

- **`packages/core/src/utils/pure.ts`** — `readErrorField(e: Error, …)` → `(e: unknown, …)`. Body already did `(e as unknown as Record<string, unknown>)`; widening the signature removes the cast at both `safeErrorToString` call sites with zero behavioral change.
- **`packages/core/src/utils/pure.ts`** — `sanitizeRankerError`: one local `const errLike = isErrorLike ? (err as Error) : null;` captures the runtime invariant; the three uses of `err.constructor`/`err.cause` are rewritten against `errLike`. Three TS18046 errors collapse to a single cast.
- **`packages/core/src/services/RetrievalService.ts`** — site 8 (phase-2 catch): one `(err as Error)` cast in the `isErrorLike ? err : new Error(...)` conditional; fixes the TS2345 on `onRetrievalFallback(error)`.

### Verification

- typecheck — pass (`pnpm -r typecheck`)
- tests — pass, **1079/1079** (`pnpm --filter @equationalapplications/core-llm-wiki test`)
- Proxy regression — pass, **12/12** (`pnpm --filter @equationalapplications/core-llm-wiki exec vitest run __tests__/instanceofErrorProxyGuard.test.ts`)
- monorepo build (includes `dts`) — pass (`pnpm -r build`)
- Test workflow on push — pass ([run 32040203500](https://github.com/equationalapplications/expo-llm-wiki/actions/runs/32040203500), green after the first attempt failed on a transient GitHub codeload 429 — re-ran clean)

No review threads were present on the PR, so nothing to resolve.

equationalapplications:
## /fix-pr follow-up #2

**Commit:** `4a91c9f`

### Resolution (hostile-Proxy surface expansion)

The earlier `/fix-pr` typecheck fix (`8f37444`) preserved the spec's `let + try/catch` idiom for `instanceof Error`. A subsequent `/code-review ultra --fix` audit surfaced three additional attacker-controlled surfaces the original Tier-A guard did not directly address. All three are now hardened, with no behavior change on benign inputs.

1. **`toString` / `Symbol.toPrimitive` traps.** After `instanceof Error` returns `false` (getPrototypeOf trap rejected), the original `sanitizeRankerError` and `RetrievalService` site-7/site-8 caught-branches still called `new Error(String(err))`. A hostile Proxy whose `toString` trap rejects would unwind that path, defeating the non-throwing contract. Now delegated to `safeErrorToString`, which converges on a static marker on throw.

2. **`err.constructor?.name` and `err.cause` property access.** Optional chaining `?.` short-circuits on null/undefined only — a hostile `Error` subclass with a throwing getter on `cause` or `constructor` still escapes. Each access in `sanitizeRankerError` is wrapped in a try/catch with a typed fallback.

3. **`sanitized.name = typeName` setter.** `Error.prototype.name` can be replaced with a hostile setter; wrapped in try/catch so the call site falls back to the default `'Error'` name.

Test coverage added in `packages/core/__tests__/instanceofErrorProxyGuard.test.ts`:

- `hostileToStringProxy` fixture (getPrototypeOf + toString/valueOf/Symbol.toPrimitive traps).
- `HostileCauseError` fixture (Error subclass with throwing `cause` getter).
- Exact-match assertions on the weakest two cases (site-8 regression + `sanitizeRankerErrors=false` wrap path), locking the observable `[object Object]` return value.
- Companion exact-match assertions in `packages/core/__tests__/vectorRanker.test.ts` site-7/site-8 integration tests, locking the same observable return value through the public `wiki.read` path.

Spec updates (`docs/superpowers/specs/2026-08-17-instanceof-error-proxy-guard-design.md`): line-number refresh for sites 2–4 (`pure.ts:592`) and site 8 (`RetrievalService.ts:513`) after the added hardening comments shifted content.

### Verification

- typecheck — pass (`pnpm -r typecheck`)
- tests — pass, **1087/1087** (`pnpm --filter @equationalapplications/core-llm-wiki test`; was 1079; +8 from new hostile-toString + HostileCauseError coverage)
- monorepo build (includes `dts`) — pass (`pnpm -r build`)
- Test workflow on push — pass ([run 32041834267](https://github.com/equationalapplications/expo-llm-wiki/actions/runs/32041834267), green after the first attempt hit a transient GitHub codeload 503/429 — re-ran clean)

equationalapplications:
@coderabbitai please review

equationalapplications:
## /fix-pr follow-up

**Commit:** `d2d2c50`

### Review resolution
- **RetrievalService.ts cause assignment (Major)** — Fixed: wrapped `error.cause = pendingRankerFallbackError` in try/catch so a hostile setter or frozen error object can't escape the catch block.
- **pure.ts constructor.name coercion (Major)** — Fixed: `typeName`/`causeName` reads now validate the result is a `string` before use, falling back to `'Error'`/`typeof` otherwise, since a hostile getter can return an object whose `toString` throws instead of throwing directly.
- **vectorRanker.test.ts duplicate assertions (Nitpick)** — Fixed: removed the repeated `toBeDefined()`/`toBeInstanceOf(Error)` lines.
- **Audit script cross-file false positives (Nitpick)** — Fixed: `has_try` now resets when the filename changes and the `try`→`instanceof Error` match is bounded to a 3-line distance.

### Verification
- typecheck — pass (`pnpm run typecheck` in packages/core)
- lint — N/A (no lint script in packages/core)
- tests — pass, 1087/1087 (`pnpm run test` in packages/core)

equationalapplications:
The /code-review high on PR #97 finished. It found no correctness bug in the TypeScript changes
  — the six guarded sites are the complete Tier-A set, the guard idiom is correct at each (try
  wraps only the operator, catch falls to a non-Error path, follow-on coercions independently
  wrapped), and readErrorField's widening to unknown is safe. Tests: 89 files / 1087 pass (PR
  body's 1079 is stale). Spec audit awk → Protected: 6/6; CI's new grep step passes locally.

  All five findings are in the docs/workflow shipped alongside. I spot-checked the two mediums
  against the files — both confirmed:

  1. README.md:125 — medium. Number(info.lastInsertid) — better-sqlite3's property is
  lastInsertRowid. undefined → NaN on every write through this adapter. It's labelled the
  canonical pattern, and the repo's two other adapter snippets (README.md:451,
  packages/core/README.md:1073) both spell it correctly, so it's a fresh typo.

  2. README.md:142 — medium. The GraphRAG end-to-end recipe goes from createWiki(adapter, …)
  straight to await wiki.ingestDocument(…) with no await wiki.setup(). createWiki only
  constructs; tables are created in WikiMemory.setup(). Copy-pasting against a fresh DB fails
  with "no such table: llm_wiki_entries". The README's six other createWiki examples all call
  setup().

  3. packages/schema-org/README.md:31 — low. The "polymorphic edges" bullet lists knows, object,
  agent — each appears exactly once in src/index.ts with a single source/target pair, so they
  aren't polymorphic. The actually-polymorphic ones are location, organizer, about, itemReviewed
  — which is what the file's own header comment says.

  4. RetrievalService.ts:528 — low. try { (error as any).cause = pendingRankerFallbackError } 
  catch {} is followed unconditionally by clearing pendingRankerFallbackError. If a plugin
  returns a frozen Error, the assignment throws, the catch swallows it, and the ranker diagnostic
  is dropped entirely rather than degraded. A second onRetrievalFallback?.(…) in the catch would
  preserve it.

  5. .github/workflows/test.yml:57 — low. The step named "Verify README GraphRAG SQL is in sync
  with EdgeRepository" only checks three substrings appear somewhere in each file. Change WHERE 
  w.depth < ? to <= and it still passes — it can't detect the drift it's named for, and it fails
  the whole test job on a docs-only reflow. Either diff the extracted fence or rename it to a
  presence check.

  Findings 1 and 2 are worth fixing before merge — they're both in the first code block a new
  user copies.

equationalapplications:
All 5 findings fixed, tests pass (1087/1087), pushed as commit d660d76, and the review summary
  is posted to PR #97.

PR #111, opened 2026-08-29
docs(spec): schema-ea executive ontology package (executive agent manifest)

equationalapplications:
## Summary

Implements [`docs/superpowers/specs/2026-08-28-schema-ea-executive-ontology-spec.md`](docs/superpowers/specs/2026-08-28-schema-ea-executive-ontology-spec.md) (rev 9, Approved → Implemented) by shipping a new data-only workspace package `@equationalapplications/schema-software-org` exporting `schemaSoftwareOrgManifest`.

This PR supersedes the earlier spec-only proposal on the same branch. The implementation lands across four code commits plus one spec-status commit:

- `d988406` — `feat(schema-software-org): scaffold package with warm-agent rows and parity guard`
- `4e74c24` — `feat(schema-software-org): add software-org types, edges, and drift guards`
- `15555f0` — `docs(schema-software-org): add package README with node and edge catalogs`
- `9c048f4` — `chore(release): wire schema-software-org into the release pipeline`
- `649b7b1` — `docs(spec): mark schema-software-org spec as Implemented`

## What the package does

A new `@equationalapplications/schema-software-org` package at `packages/schema-software-org/`. Data-only; no runtime logic. Exports one `OntologyManifest` literal — `schemaSoftwareOrgManifest` — with **17 node types** and **40 edge types**:

- **9 warm-agent types** copied **verbatim** from `packages/schema-org/src/index.ts` (D2). The single sanctioned divergence is on `product` (D8 — drops "and software" from the upstream text and hands off to `software_application` / `service`).
- **3 software-org base types** — `software_application`, `service`, `role`. No parent, no children (D5).
- **5 `creativework` subtypes** — `design_spec`, `handoff`, `procedure`, `session_recap`, `reference_doc`. One-level polymorphic inheritance via `parent_type` (D6).
- **40 edges** — 28 warm-agent rows copied verbatim plus 12 new software-org edges. `supersedes` declared once as `creativework → creativework` (D6 — parent-aware matching on both sides).

All properties live in each node's `description` string (D3), so no code changes to `core-llm-wiki` are required. Serialized manifest is **11,521 bytes** (in the spec's 11,000–12,000 byte window).

## Verification

- 34/34 tests pass in `packages/schema-software-org/`:
  - 12 warm-agent parity tests (D2 guard; fails on byte-level drift, including order)
  - 7 shape/counts tests (D5: parent_type resolves to top-level only)
  - 8 D3 property-convention tests (`it.each` across the 8 new node types)
  - 2 D6 shape guards (supersedes declared once, on the parent)
  - 4 D8 disambiguation-text guards
  - 1 content-drift snapshot test
- Workspace-wide `pnpm install` / `pnpm run build` / `pnpm run typecheck` / `pnpm test` all pass — 1,592 tests across 11 packages, 0 failures.

## Files added / modified

**New:**
- `packages/schema-software-org/{package.json,tsconfig.json,tsup.config.ts,vitest.config.ts,LICENSE}`
- `packages/schema-software-org/src/index.ts` (the manifest)
- `packages/schema-software-org/__tests__/manifest.test.ts`
- `packages/schema-software-org/__tests__/__snapshots__/manifest.test.ts.snap`
- `packages/schema-software-org/README.md`

**Modified:**
- `.releaserc.json` — added `schema-software-org` to the `prepareCmd` package list so its version gets bumped each release.
- `.github/workflows/release.yml` — added a `publish_if_needed` line **after** the `core` line (load-bearing dependency order).
- `README.md` (root) — added the npm badge row and a row in the package table.

**Spec:**
- `docs/superpowers/specs/2026-08-28-schema-ea-executive-ontology-spec.md` — Status: **Implemented** (separate commit per repo convention).

## Dependency / install note

`@equationalapplications/schema-org-llm-wiki` is a **devDependency** only (used by the parity test). It does **not** appear in `dependencies` — this package is a self-contained superset of the warm-agent manifest.

Runtime requires `@equationalapplications/core-llm-wiki >= 6.1.0` (the release that shipped `OntologyNodeType.parent_type`).

## Before merging — npm trusted publishing prerequisite

Configure npm trusted publishing for `@equationalapplications/schema-software-org` on npmjs.com. Publishing is OIDC-based; an unconfigured package name fails at publish time, after the release tag has already been created. This cannot be done from the repo — it's a manual step on npmjs.com.

## Out of scope (optional, reviewer-requested)

Task 5 in the plan — an end-to-end `design_spec → design_spec` `supersedes` round-trip through `WikiMemory` — was marked optional and **skipped** by default. Core's own `ontologyParentInheritance.test.ts` plus the manifest's Test 8 (`supersedes stays declared on the parent`) already cover the parent-matching mechanism this would exercise. Happy to add it as a follow-up commit if the reviewer wants the combination proven end-to-end.

## Rulings made during execution

- **Test-count undercount in the plan.** Plan said "expect 21" (12 prior + 9 new). The actual count is 34 because the plan counted `it.each` blocks as one test rather than expanding them. Every behavior the plan listed is covered; the count discrepancy is purely `it.each` accounting.
- **`pnpm test -- -u` does not forward `-u` through the `vitest run` script.** Used `pnpm exec vitest run -u` directly for snapshot regeneration. Tooling note, not a spec deviation.
- **`product` D8 override text must be byte-identical in `src/index.ts` and the test's allowlist.** Verified by the parity test's `expect(copied!.description).toBe(override)` and the override-staleness guard.


equationalapplications:
## /fix-pr follow-up

**Commit:** `58627f5`

Note: this PR now contains three spec revisions beyond the reviewed state (rev 2, 3, 4). The review threads below are resolved against rev 4.

### Review resolution

- **Warm-agent edge descriptions abbreviated in the catalog table** (`PRRT_kwDOSP9TPs6dg_tT`, line 226) — **Fixed.** Correct and still live. Rev 2 restored the manifest code block to verbatim but left 10 of 28 edge rows abbreviated in the catalog *table*, so the document defined two values for one row — the exact defect rev 2 set out to fix. Rewritten from `packages/schema-org/src/index.ts`: `containedInPlace`, `subOrganization`, `object`, `about` (→place, →event), `itemReviewed` (→creativework, →organization, →place, →product), `owns`. Verified table, manifest block, and upstream agree on all 28 rows; node table matches too (`product` excepted, D8). D2 now holds the tables to the same standard as the manifest.

- **Implement and test inherited ontology behavior** (`PRRT_kwDOSP9TPs6dZZzZ`, outdated) — **Partly fixed in spec; core implementation not applied here.** The finding was directionally right but misattributed on both halves:
  - *Prompt serialization.* No core change needed. `OntologyService.buildPromptContext` does `JSON.stringify(manifest, null, 2)`, so description text — property list included — already reaches the classification prompt verbatim. Nothing is expanded because nothing needs expanding. The real defect was the spec's claim that subtypes *inherit* properties from the parent: false, and doubly so since `creativework` declares no properties to inherit. Removed in rev 2 as **D4** (there is no property inheritance; each concrete type carries its own complete list).
  - *Edge resolution.* Real, but the failing side is the **target**, not the source. Source-side parent matching lands in #110. Target matching stays exact (`OntologyService.resolveEdges:141` compares `def.target_type` to the target fact's `okf_type`), so the `supersedes` row declared as `creativework → creativework` could never resolve a `design_spec → design_spec` pair — the edge was dead on arrival. Fixed data-side in rev 2 by enumerating one row per concrete type (**D6**), edge count 40 → 45. The bot's `author` example is not affected: `author` is declared on `creativework` as *source*, which #110's source-side matching covers.
  - No core changes made: this is a spec-only PR, and the source-side work belongs to #110.

### Also addressed in this PR since review

- `parent` → `parent_type` throughout (dependency spec #110 rev 3; the field collides with the warm-agent `parent` edge type).
- Node count corrected 18 → **17** (3 EA base types, not 4) — resolves the open question in the PR description's review checklist.
- Warm-agent node/edge rows restored verbatim (rev 1 abridged all 37 ~50% while claiming verbatim) with a devDependency parity test as the drift guard (**D2**).
- `product` / `software_application` / `service` overlap resolved inside the manifest descriptions rather than by convention (**D8**), with a reasoned override allowlist in the parity test.
- "Parent Types" → "Base Types"; package skeleton, core version pin, MD040 tag.

### Verification

- typecheck — pass (`pnpm -r typecheck`, after `pnpm build:core`; the first run failed on a stale `packages/core/dist`, unrelated to this change)
- lint — N/A (no lint script in root or package `package.json`)
- tests — pass (`pnpm -r --filter='!@equationalapplications/integration-llm-wiki' test`, plus integration: 7 files / 41 tests)

Docs-only change; no runtime code touched.

equationalapplications:
## Spec rev 5

**Commit:** `c054acf`

Follows #110 rev 4 (`763d720`), which made type matching parent-aware on **targets** as well as sources, resolved exact-first. That removes the reason rev 2 enumerated `supersedes`:

- **`supersedes` collapses 6 rows → 1.** The single `creativework → creativework` row now resolves a `design_spec → design_spec` pair.
- **The `itemReviewed → creativework` limitation is gone.** Review edges reach all five concrete subtypes natively.
- **Edge count 45 → 40** (28 warm-agent + 12 EA).
- **D6 rewritten** from "target matching is exact — enumerate the pairs you need" to "declare on the parent". The accepted cost stands: a `→ creativework` target admits every child, with an explicit concrete row as the escape hatch, since the exact-first pass makes it win.

Verified the catalog table, the manifest block, and `packages/schema-org/src/index.ts` agree on all 40 rows.

equationalapplications:
## /fix-pr follow-up

**Commit:** `a0b1ab8`

### Review resolution
- **MD028 blank line inside blockquote (line 509):** **Fixed.** Replaced the blank line with `>` so the blockquote stays continuous. (file: `docs/superpowers/specs/2026-08-28-schema-ea-executive-ontology-spec.md`)
- **Verbatim-superset claim should note the `product` override (lines 3 & 129):** **Fixed.** Updated both locations in `packages/schema-software-org/README.md` to state warm-agent rows are verbatim except for the intentional `product` override (which delegates to `software_application` / `service`).
- **Monorepo package count says seven but the table has eight rows (line 230):** **Fixed.** Changed "seven packages" to "eight packages" in the root `README.md`.

### Verification
- lint — N/A (no `lint` script in any package)
- tests — pass (`pnpm test`): schema-software-org 34/34, expo 10/10, react 73/73, prisma-outbox 15/15, schema-org 6/6, scopelab 11/11, wiki-demo 5/5, integration 41/41; workspace build also passes (`pnpm run build`).

equationalapplications:
This is an incredibly impressive and rigorous audit. The 14-agent roster pipeline paid for itself entirely by catching that `release.yml` republish bug—that is a classic monorepo footgun that would have caused a very confusing CI failure the next time you needed to backport a hotfix to a `v5.x` or `v4.x` tag.

Here is the blueprint for resolving the requested changes so we can get this PR green and merged.

### 1. The Critical Fix: `release.yml` existence guard

You are completely right. The `publish_if_needed` bash function in `.github/workflows/release.yml` blindly assumes the package `package.json` exists, which breaks `workflow_dispatch` on older tags.

**The Fix:** We will add a standard `[ -f ]` file existence guard at the top of the `publish_if_needed` function definition.

```yaml
publish_if_needed() {
  local package_json="$1"
  shift
  # Guard against missing packages on older tag checkouts
  if [ ! -f "$package_json" ]; then
    echo "Skipping: $package_json not found (likely an older tag checkout)."
    return 0
  fi
  
  # ... existing node -p require(...) logic ...
}

```

### 2. Suggestion R1-7: Two-way Parity Guard in `manifest.test.ts`

Excellent catch on the mutation test. The current filter just plucks the upstream rows out of the new manifest and compares them, silently leaving behind any rows that upstream deleted.

**The Fix:** Add a strict length assertion on the total edges to ensure no stale rows are hiding in the array. Since we know exactly how many new edges we added (12), the math must always balance.

*Add to `manifest.test.ts` (around line 66):*

```typescript
// Reverse audit: ensure no stale rows were left behind if upstream removed them.
// The total must exactly equal the upstream count plus our 12 new edges.
expect(schemaSoftwareOrgManifest.edge_types).toHaveLength(
  schemaOrgWarmAgentManifest.edge_types.length + 12
);

```

### 3. Suggestion R1-4: The 6.0.1 Validation Claim

Good catch on the core behavior. `validateManifest` allows unknown fields, so it won't throw; it will just drop `parent_type` and fail silently later when the classification LLM tries to route an edge.

**The Fix:**

* **In `packages/schema-software-org/README.md:127`:** Change "This manifest does **not** validate against 6.0.1" to *"This manifest will validate against 6.0.1, but will silently fail at ingest edge matching because 6.0.1 ignores the `parent_type` field."*
* Update the corresponding claim in the spec.

### 4. Suggestion R1-3: Spec `tsconfig` claims

Accuracy matters in the historical record.

**The Fix:** Edit the spec (lines 296/357) to remove the claims that `tsconfig.json` "extends monorepo root" and that there are "no project references." Update the reasoning to state: *"Uses a standalone configuration mirroring `schema-org`, with no changes needed."*

### 5. Suggestion R1-1: Monorepo Ecosystem Tables

A classic documentation omission.

**The Fix:**

* Add the `## Monorepo Ecosystem` table (complete with all 8 packages) to the bottom of the new `packages/schema-software-org/README.md`.
* Propagate the updated 8-package table to the `README.md` files of the other 7 packages (`core`, `react`, `expo`, `prisma-outbox`, `core-llm-tools`, `okf`, `schema-org-llm-wiki`) so they are all in sync.
* Update the spec's Monorepo Integration checklist to include this step.

equationalapplications:
## /fix-pr follow-up

**HEAD:** `8fb950c16eb15474129a64c720031be11a68cfb6`

### Review resolution

Five fixes from the blueprint, landed as three reviewable commits:

- **R1-crit: `release.yml` republish footgun** — **Fixed** in `dce4458e`.
  `publish_if_needed` now skips cleanly with a diagnostic `return 0` when
  the package.json is absent (workflow_dispatch on older tags). Files:
  `.github/workflows/release.yml`.

- **R1-7: Two-way parity guard** — **Fixed** in `64a4c1ad`. Added a
  reverse-audit length assertion (`upstream_count + 12`) so a stale edge
  or a missed upstream removal surfaces as a count mismatch instead of a
  silent test pass. Files:
  `packages/schema-software-org/__tests__/manifest.test.ts`.

- **R1-4: 6.0.1 validation claim** — **Fixed** in `64a4c1ad`. The README
  copy now says the manifest *will* validate against 6.0.1 but silently
  fails at ingest edge matching because 6.0.1 ignores `parent_type`.
  Files: `packages/schema-software-org/README.md`.

- **R1-3: Spec `tsconfig` claims** — **Fixed** in `8fb950c1`. Removed
  the claim that `tsconfig.json` "extends monorepo root" and the trailing
  "(no project references, no `paths`)" sentence; replaced with an
  accurate statement that `packages/schema-software-org/tsconfig.json`
  is a standalone config mirroring `schema-org` and is intentionally
  absent from the root tsconfig `references` array. Files:
  `docs/superpowers/specs/2026-08-28-schema-ea-executive-ontology-spec.md`.

- **R1-1: Monorepo Ecosystem tables** — **Fixed** in `8fb950c1`.
  Added the missing 8th row to all seven existing per-package
  `## Monorepo Ecosystem` tables (`core`, `expo`, `react`, `prisma-outbox`,
  `core-llm-tools`, `okf`, `schema-org`), and added a freshly-built 8-row
  table to the new `schema-software-org` README. Spec's Monorepo
  Integration checklist now has a 5th item calling this out.
  Files: 8 package READMEs + the spec.

### Verification

- typecheck — pass (`pnpm run typecheck`, all 11 workspace projects clean)
- lint — N/A (no lint script in this repo)
- tests — pass (`pnpm test`, 1,544 tests across 11 workspace packages + 41 integration tests; new reverse-parity test makes `schema-software-org` 35/35)

PR #114, opened 2026-09-02
feat(core): atomic multi-entity manifest seeding via setOntologyManifests

equationalapplications:
Add `WikiMemory.setOntologyManifests` for atomic, conflict-safe seeding of several ontology manifests in one transaction, and make the singular `setOntologyManifest` delegate to it. `opts.ifAbsent` switches the underlying upsert to `ON CONFLICT DO NOTHING`, closing the TOCTOU window that makes two concurrent initializers silently overwrite each other's manifests.

See [`docs/superpowers/specs/2026-09-02-atomic-multi-entity-manifest-seed-design.md`](docs/superpowers/specs/2026-09-02-atomic-multi-entity-manifest-seed-design.md) for design and rationale.

- New public method `setOntologyManifests(entries, opts?)` returning `{ written, skipped }`
- `MetadataRepository.setManifest` gains `opts.ifAbsent` and reports whether it wrote
- Input validation (duplicates, manifest shape) runs before the transaction opens
- Singular `setOntologyManifest` delegates; observable contract unchanged
- README documents the method and the `ifAbsent` interaction with `seedManifests`
- 12 new tests in `__tests__/setOntologyManifests.test.ts`; 6 added to the existing repository test

No schema migration. No new outbox event. Singular method's signature, return type, and semantics unchanged.

https://claude.ai/code/session_0187MubfgYcykMfyoLb2zXwD

### DIFF
--- a/docs/superpowers/specs/2026-09-02-atomic-multi-entity-manifest-seed-design.md
+++ b/docs/superpowers/specs/2026-09-02-atomic-multi-entity-manifest-seed-design.md
@@ -2,6 +2,9 @@
 
 **Date:** 2026-09-02
 **Status:** Implemented (2026-09-02) — branch `spec/atomic-multi-entity-manifest-seed`
+**Revised:** 2026-09-02 — review found §3 assumed `entity_manifests` is the only
+source of a manifest; it is not. See *`ifAbsent` and config-seeded manifests* under
+Cross-cutting. Documentation-only amendment; no code changed.
 **Issues:** none filed; raised from consumer review (see Consumer context)
 **Packages:** `@equationalapplications/core-llm-wiki`
 **Consumer context:** `equationalapplications/curated-thoughts` — spec
@@ -270,6 +273,28 @@ This deletes its `getOntology` pre-read loop, its compensating-rollback block,
 and its local empty-manifest constant. Its never-throws contract is preserved by
 catching and reporting failure; it no longer has to *undo* anything, because the
 engine rolled back.
+
+**`ifAbsent` and config-seeded manifests.** §3 defines "already present" as a row in
+`entity_manifests`, and the implementation is faithful to that. But a manifest can
+also come from `WikiConfig.ontology.seedManifests`, and that route persists lazily:
+`OntologyService.getEffectiveState` writes the seed to the table only when resolved
+with a `tx` (ingest, heal); resolved without one it caches in memory, and
+`WikiMemory.getOntologyManifest` returns the seed without persisting it. So for an
+entity configured that way, `ifAbsent` reports `written` and replaces the seed
+before any ingest has run, and reports `skipped` after — the same call, the same
+config, opposite outcomes depending on unrelated activity. A consumer that
+pre-checked with `getOntologyManifest` would see a manifest "present" and then watch
+`ifAbsent` overwrite it.
+
+This does not affect the motivating consumer, which seeds through the API and not
+through `seedManifests`, and it is not worth closing in code: making `ifAbsent`
+consult the config seed would mean the batch could skip an entity that has no row,
+leaving the "seeded" state observable only through configuration — a worse
+invariant than the one it replaces. The rule is instead a documented one: seed an
+entity through the config or through this method, not both. `packages/core/README.md`
+states it at the point of use, alongside a correction to its claim that
+`seedManifests` entries are written "on first access" (only tx-scoped access
+persists them).
 
 **A caveat the consumer must record.** That consumer seeds its two fixed tiers
 during startup and its workspace tier later, once the workspace id resolves —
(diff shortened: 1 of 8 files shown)


equationalapplications:
## /fix-pr follow-up

**Commit:** `371dd18`

### Review resolution
- **CodeRabbit nitpick — markdownlint MD040 on `docs/superpowers/specs/2026-09-02-atomic-multi-entity-manifest-seed-design.md` line 208** — **Fixed:** added `text` language identifier to the `WikiMemory.setOntologyManifests` pseudocode fence so MD040 passes. (`docs/superpowers/specs/2026-09-02-atomic-multi-entity-manifest-seed-design.md`)
- **CodeRabbit pre-merge warning — Docstring coverage 75% < 80%** — **Fixed:** added a JSDoc to the new `makeWiki` test helper in the batch test file, bringing the four touched-function count (3/4 → 4/4) over the 80% gate. (`packages/core/__tests__/setOntologyManifests.test.ts`)
- **Other CodeRabbit review items** — none outstanding; the PR was APPROVED with no other review threads, and `aws-cloud-agent-pr-review` reported a transient backend failure rather than actionable feedback, so it is intentionally not addressed here.

### Verification
- typecheck — pass (`pnpm --filter @equationalapplications/core-llm-wiki typecheck` → `tsc --noEmit`, clean)
- lint — N/A (no lint script defined for `packages/core`; CodeRabbit MD040 is the only linter that flagged this PR and is fixed)
- tests — pass (`pnpm --filter @equationalapplications/core-llm-wiki test` → 1193 passed across 92 files; the two affected files, `setOntologyManifests.test.ts` and `repositories/MetadataRepository.manifest.test.ts`, run 24/24 green)

equationalapplications:
Re the Copilot review on the `ifAbsent` concurrency test — the diagnosis was correct and is now fixed in f47c6de, though by a different route than suggested.

**The diagnosis holds.** I confirmed it by mutation rather than inspection: rewriting `MetadataRepository.setManifest`'s `ifAbsent` path as a check-then-write (read first, then a plain upsert) left all 17 tests passing. The suite genuinely could not distinguish the atomic `INSERT ... ON CONFLICT DO NOTHING` from the TOCTOU-prone form that `ifAbsent` exists to rule out. Both calls went through one `WikiMemory`, so the per-instance mutex in `withSerializedTransactions` (`WikiMemory.ts:89`) serialized them.

**The suggested fix isn't viable here, and would assert something false.** The suggestion was separate instances against the same database, asserting the loser "neither overwrites nor errors." I built exactly that. The loser *errors*: `WikiTransactionError: Transaction failed`. The mutex is per-instance, so two instances issue overlapping `BEGIN`s on the one shared better-sqlite3 connection and the second is rejected. That assertion would fail against correct code. Separate connections aren't an option either — each `new Database(':memory:')` is a distinct empty database, not a shared one.

**What replaced it.** The race isn't reproducible on this harness, but the property that makes a race safe is: the write consults no prior read. The new test `ifAbsent decides the conflict in a single statement, with no prior read` seeds a row, then counts `SELECT`s against `entity_manifests` during a conflicting `ifAbsent` write and asserts zero — a check-then-write *must* read there to learn it should skip. Re-running the same mutation, it now fails, and it is the only failure.

The misleading test was also renamed to `two overlapping ifAbsent batches on ONE instance converge, neither erroring`, with a comment noting it pins idempotent convergence rather than a cross-instance race, and pointing at the atomicity test. The original name did oversell what it verified.

1246/1246 tests passing, `tsc --noEmit` clean.

https://claude.ai/code/session_011bL2yMkNEi1DkY7ZwD5Hq2

equationalapplications:
## /fix-pr follow-up

**Commit:** `bec28439b4ebd7e77dd3487defe64d99cc606e59`

### Review resolution
- **Unresolved review threads:** none — 1 thread total on this PR, already resolved. No review feedback required action.
- **Failing check `test`:** — **Fixed:** the failure was in the *Audit (fail only on advisories with an upstream patch)* step, not the unit tests. Four new high-severity `fast-uri` advisories (`GHSA-5jgf-p345-68v8`, `GHSA-f65p-4m7j-42xc`, `GHSA-fph4-wmhf-6fwf`, `GHSA-jqff-g426-hqxp`) are patched in `>=3.1.6`, but the existing override capped at `3.1.5` — a pin written for an earlier advisory round, and now itself vulnerable. Bumped the override to `3.1.6` (files: `pnpm-workspace.yaml`, `pnpm-lock.yaml`).

Notes:
- Unrelated to this PR's `setOntologyManifests` changes. The prior commit `43d78ff` passed 9 minutes earlier and the failing commit `f47c6de` was test-only, so the advisories landed in the DB between the two runs.
- `fast-uri` is a dev-only transitive of `ajv@8.20.0`, which is why the earlier `--prod` allowlist audit step stayed green and only the full-tree classifier step failed.
- Lockfile delta is scoped to `fast-uri` 3.1.5 → 3.1.6 only; no other resolutions moved.
- The inert `fast-uri@>=2.0.0 <3.0.0: 2.4.4` override was left alone — no 2.x is present in the tree.

### Verification
- audit gate — pass (`pnpm audit --audit-level=high --json` + the workflow's classifier script, reproduced locally; only the two known-unpatched `image-size` warnings remain)
- build — pass (`pnpm run build`)
- tests — pass (`pnpm test`, exit 0; core 1195 passed / 92 files, integration 41 passed, plus react/expo/schema/apps suites)
- `setOntologyManifests.test.ts` — pass (18 tests)
- lint — N/A (no `lint` script in the workspace)

Issue #118, opened 2026-09-03
embedFact failures are not retryable: no persisted marker distinguishes recoverable failure from 'no embed provider'


equationalapplications:
## Summary

Deferred follow-up called for in §9 of the tierFloors + embed clamp design spec, split out of #104 so that issue could close on the clamp alone.

`EmbeddingService.embedFact` returns `false` and logs a `console.warn` on **every** failure mode — over-length input, provider outage, malformed or non-finite vector, dimension mismatch. Nothing is persisted to record *that* the fact failed to embed, or *why*.

## Why it matters

`runReembed` cannot distinguish:

- "never embedded because the provider was down" (recoverable — should retry)
- "never embedded because no `embed` is configured" (not recoverable — retrying is pure waste)
- "never embedded because the input was rejected as too long" (recoverable only after the clamp change in #104)

All three present identically: `embedding_blob IS NULL`. So `runReembed` either retries everything on every pass, or cannot target retries at all. A fact that fails once during a transient provider outage stays unembedded and invisible to semantic retrieval indefinitely, with the only trace a warning line in whatever captured stdout at the time.

The clamp landed in #104 removes one *cause* of silent permanent failure but not the underlying gap: there is still no durable record of a failed embed.

## Scope

Per spec §6.4, this was explicitly held out of the #104 change because it touches three things the clamp did not:

1. **Persistence** — a marker for "embedding attempted and failed", ideally with a coarse reason and an attempt count / last-attempt timestamp. Needs a schema addition and migration.
2. **`embedFact`'s return contract** — `Promise<boolean>` cannot express "failed, retryable" vs "failed, terminal" vs "no provider". Widening it affects every caller.
3. **`runReembed` orchestration** — selecting retry candidates, applying backoff, and deciding when a fact is permanently un-embeddable.

## Suggested direction

Write a short spec first; the return-contract change in particular is a public-API decision. Options worth weighing there: a nullable `embedding_failed_at` / `embedding_failure_kind` pair on `llm_wiki_entries` versus a separate failures table, and whether `embedFact` should return a discriminated result object or keep `boolean` and surface failure detail through a hook.

## Context

- Spec §9 and §6.4: `docs/superpowers/specs/2026-09-02-tier-floors-and-embed-clamp-design.md`
- Current behavior: `packages/core/src/services/EmbeddingService.ts` (`embedFact`, the `catch` that warns and returns `false`)
- Split from #104

[...]

PR #119, opened 2026-09-03
feat: tierFloors and configurable embed input clamp

### LINKED ISSUE #104: EmbeddingService.embedFact has no input length guard — oversized facts silently skip embedding
## Description
`EmbeddingService.embedFact()` (packages/core/src/services/EmbeddingService.ts) passes `${title} ${body} ${tags}` to `llmProvider.embed()` with no check against the embedding model's context window. If an LLM-generated fact body exceeds the embed model's token limit (e.g. nomic-embed-text = 2048 tokens ≈ ~8k chars), the provider errors and the fact is silently left without an embedding (caught + `console.warn` only).

## Context / evidence
Found while debugging the same failure class in curated-thoughts: Ollama returns HTTP 400 "the input length exceeds the context length" for over-length inputs. Verified directly against local Ollama with a 15k-char document.

## Proposed fix
- Clamp/truncate or split the embed input in `embedFact` to a configurable max (config knob, e.g. `maxEmbedChars`, defaulting safely under common embed-model limits), OR
- Surface failed embeddings as a retryable state instead of silently dropping.

## Notes
- The primary fix for the actual incident lives in curated-thoughts' Rust pipeline (it embeds raw chunks unguarded); this issue covers defensive hardening of the TS packages' fact-embedding path.
- Related: `chunkText` in IngestionService defaults maxChunkLength=12000, but those chunks go to fact extraction (LLM), not the embedder.

### LINKED ISSUE #109: ReadOptions: per-entity result floor (tierFloors) — a higher-weighted entity can starve another out of the bundle entirely
## Problem

`ReadOptions.tierWeights` controls *relative ranking* between entities in a multi-entity
read, but there is no way to guarantee that a given entity contributes **any** results at
all. When one entity is both weighted higher and much larger than another, it can take every
slot within `maxResults`, and the lower-weighted entity is silently absent from the bundle.

This is not a tuning problem — no weight assignment fixes it, because the failure is a
function of the *size ratio* between entities, which drifts as data accumulates.

## How it bit us

In `aws-cloud-agent` a code-review agent reads three tiers in one call:

```ts
wiki.read(
  ['guidance@owner/repo', 'codebase@owner/repo', 'business'],
  seedQuery,
  { maxResults: 25, hybridWeight: 0.7, tierWeights: {
      'guidance@owner/repo': 2.5,
      'codebase@owner/repo': 2.0,
      'business': 0.25,
  }},
);
```

`guidance` is weighted above `codebase` deliberately: when the bundle is capped, a stated
rule should survive at the expense of a structural fact. That is correct for the *marginal*
slot.

It stopped being correct once the tiers diverged in size:

| entity | live facts | weight |
| --- | --- | --- |
| `guidance@…/aws-cloud-agent` | 6,655 | 2.5 |
| `codebase@…/aws-cloud-agent` | 1,841 | 2.0 |

At a 3.6:1 ratio, guidance facts fill all 25 slots and the returned bundle contains **zero**
codebase facts — against a fully populated codebase tier. The consumer requires codebase
context to do its job, so it correctly refuses to proceed, and the review is skipped.

Measured over 7 days: 17 of 161 review runs (11%) were abandoned this way. The rate tracks
the guidance:codebase ratio across repositories — 3.6:1 → 5 occurrences, 2.2:1 → 1,
0.8:1 → 2 — which is what identified the cause.

## Proposed API

A per-entity floor alongside `tierWeights`:

```ts
interface ReadOptions {
  // …
  /**
   * Minimum number of results to retain from each named entity before the global
   * `maxResults` cut. Applied after scoring: the top-N scored results from each named
   * entity are reserved, then remaining slots fill by score as usual.
   * An entity with fewer than N matching facts contributes what it has.
   * Floors summing above `maxResults` are an error.
   */
  tierFloors?: Record<string, number>;
}
```

Our case becomes `tierFloors: { 'codebase@owner/repo': 8 }` — guidance keeps its priority for
the other 17 slots, and the bundle can never be structurally blind.

### Why it belongs in the library

The floor is a ranking concern, and the library owns ranking. A caller can only approximate
it by issuing a second, entity-scoped read and merging by score — which duplicates the
scoring the library just did, and needs `factScores` to merge correctly. We are shipping
exactly that as an interim workaround, and would rather delete it.

### Semantics worth pinning down

- Interaction with `preFilterLimit`: the floor should apply after pre-filtering, so a floor
  cannot resurrect a fact the pre-filter excluded.
- Interaction with `includeZeroWeightEntities`: a floor on a zero-weight entity is most
  likely a caller error and could reasonably throw.
- Deterministic tie-breaks presumably reuse the existing
  `score → access_count → updated_at → id` ordering.

## Environment

`@equationalapplications/core-llm-wiki` 6.0.0. `ReadOptions` verified in
`dist/testing-DgqVB29I.d.ts:228-250`: `maxResults`, `preFilterLimit`, `hybridWeight`,
`tierWeights`, `includeZeroWeightEntities` — no floor or quota field.

Happy to open a PR if the shape above looks right.

### LINKED ISSUE #117: tierFloors silently unsatisfied for facts an external vectorRanker does not score
## Summary

`ReadOptions.tierFloors` (added in #109) is silently unsatisfied when a `vectorRanker` is configured and the floored entity's matching facts are ones the ranker returns no `semanticScore` for — typically facts with no stored embedding.

The built-in JS-cosine path honors the floor on the identical corpus, so this is a path-dependent inconsistency: a host that adopts a `vectorRanker` plugin silently loses a guarantee it had before.

## Why it matters

This is the same failure class `tierFloors` was introduced to eliminate. Per the design spec §1.3, the original bug was dangerous precisely because *nothing errors* — the bundle is well-formed and simply structurally blind. That property is reproduced here: no throw, no hook, no `onRetrievalFallback`. A host that added a floor to guarantee a tier's presence gets a bundle without it and no signal.

## Cause

On the `vectorRanker` path, `RetrievalService.read()` calls the ranker **per entity** with `limit: Math.max(maxResults * 2, maxResults + 50)`, so rows the ranker *does* score are generously oversampled per entity — floors work correctly for those (verified below).

Rows the ranker **omits** are re-introduced by the backfill block. That budget is computed globally:

```ts
const maxBackfill = isHybrid ? maxResults : Math.max(0, maxResults - scored.length);
```

and filled by a single global top-K selection ordered by `_compareScoredRows` across *all* entities, with no per-entity reservation. A floored entity whose rows are all in that omitted set therefore competes globally and can win zero slots.

Note this is *not* the `preFilterLimit` interaction documented in spec §3.5, and not the data-dependent shortfall §4.3 declares acceptable — the rows exist, survive pre-filtering, and are present in `candidateRows` at the time of the cut.

## Reproduction

`maxResults: 5`, `tierFloors: { small: 2 }`, corpus of 8 `big` + 4 `small` facts, query `"shared"`:

| Ranker | `small` rows | Floor honored |
| --- | --- | --- |
| `vectorRanker` scores both entities | embedded | ✅ 2 of 2 |
| `vectorRanker` returns `[]` for `small` | un-embedded | ❌ **0 of 2** |
| none (js-cosine) | un-embedded | ✅ 2 of 2 |
| none (js-cosine) | embedded | ✅ 2 of 2 |

Only the second row fails. Result ids in the failing case are `big0..big4` — `small` is entirely absent despite a floor of 2 and four eligible facts.

```ts
const omittingRanker: VectorRanker = {
  rankBySimilarity: async (args) => {
    if (args.entityId === 'small') return [];
    return Array.from({ length: 8 }, (_, i) => ({ id: `big${i}`, semanticScore: 0.9 - i * 0.01 }));
  },
};
// 8 `big` facts WITH embeddings, 4 `small` facts with embedding_blob = NULL
const result = await wiki.read(['big', 'small'], 'shared', {
  maxResults: 5,
  tierFloors: { small: 2 },
});
result.facts.filter(f => f.entity_id === 'small').length; // 0, expected 2
```

## Suggested direction

Make the backfill budget floor-aware: reserve each floored entity's shortfall (`floor - alreadyScoredForEntity`) inside `maxBackfill` before the global top-K selection consumes it. That keeps the existing O(N log K) selection for the remainder.

This touches the ranker backfill subsystem rather than the two cut sites the original spec addressed, and changes how `maxBackfill` is apportioned, so it likely deserves its own short spec.

## Meanwhile

Documented as a known limitation on the `tierFloors` bullet in `packages/core/README.md`: keep embeddings current for floored entities, or leave `vectorRanker` unset.

## Context

- Design spec: `docs/superpowers/specs/2026-09-02-tier-floors-and-embed-clamp-design.md`
- Feature landed in #109; relevant code in `packages/core/src/services/RetrievalService.ts` (backfill block) and `packages/core/src/readOptions.ts` (`selectWithFloors`)

https://claude.ai/code/session_01KruECcCobwmGonyiaW4FHp

equationalapplications:
## Summary

Implements `tierFloors` (per-entity result floors in multi-entity reads) and `maxEmbedChars` (configurable embed input clamp) for `@equationalapplications/core-llm-wiki`.

- **#109** (tierFloors): A multi-entity read with a 3.6:1 size ratio between entities was silently starving the smaller tier when `tierWeights` alone couldn't guarantee a minimum contribution. Added opt-in per-entity result floors that reserve a top-N before the global `maxResults` cut.
- **#104** (embed clamp): Embedding input was hardcoded to 16,000 characters, which exceeds common model windows (e.g. nomic-embed-text's 2,048 tokens). Lowered default to 6,000 with a configurable knob, retaining the 16,000 hard ceiling as a security control.

Both features are independent, land as separate commits (`feat` and `fix` for semantic-release changelog), and ship in one PR per the design spec.

### `tierFloors`

- **Task 1** — Added types: `ReadOptions.tierFloors`, `MemoryBundle.metadata.tierFloors`, `WikiConfig.maxEmbedChars`, `WikiInvalidReadOptions` error class
- **Task 2** — `validateTierFloors(entityIds, tierFloors, sanitizedTierWeights, includeZeroWeightEntities, maxResults)` validates and sanitizes, throws on by-construction contradictions
- **Task 3** — `selectWithFloors<T>(sortedRows, floors, maxResults)` reserves each entity's top-N in rank order
- **Task 4** — Integrated into `RetrievalService`: validation gated on `exposeMetadata`, applied at both scored cut sites (ranker + keyword-fallback), exposed in metadata
- **Task 5 (review fix, #117)** — Closed vectorRanker path floors gap: ranker backfill path now reserves each floored entity's shortfall from the omitted pool before the global budget is spent; external-ranker keyword-fallback path now applies floor-aware oversampling limit widening

### `maxEmbedChars`

- **Task 6** — Created `DEFAULT_MAX_EMBED_CHARS = 6_000` and `EMBED_CHARS_CEILING = 16_000` exports; `embedFact` resolves config with ceiling enforcement
- **Task 7** — Documented both options in README files with defaults and behavior

### Deviations recorded in ledger

1. **Index-sort vs `_compareScoredRows` re-sort** (spec §3.4 step 3 → plan §2, Task 3): Sorting selected rows by original input index is provably equivalent for already-sorted input and avoids duplicating the private tie-break comparator.
2. **`jsCosineNeedsTierSort` widening** (spec §2 analysis gap, Task 4): The JS-cosine ranker's `limit` pre-truncates candidates before `selectWithFloors` sees them. Now fires when any floor > 0 to force full materialization.
3. **Test harness MiniSearch sync** (spec harness gap, Task 4): Raw SQL inserts bypass the write API; added `__testAccess.searchService.sync()` for keyword-fallback tests.
4. **Ranker backfill reservation** (Task 5, review fix): Computed per-entity floor shortfalls from ranker output and pre-reserves omitted rows per floored entity before the global budget is spent (pure-semantic: scored with -2; hybrid: scored with `(1-weight)*kwScore`).

### #118: Non-retryable embedding failures
`embedFact` returns `false` and logs a warning on all failure modes (provider outage, malformed vector, input over original 16K limit). No persisted marker distinguishes recoverable failures ("was down", "input now under 6K clamp") from terminal ones ("no provider configured"). `runReembed` cannot target retries. Deferred to spec §9.

**Status**: #117 (vectorRanker backfill gap) is closed by this PR (commit 86fa686).

## Verification

- **Type safety**: `pnpm typecheck` clean across all 11 packages
- **Tests**: 1235/1235 passing in core package; full monorepo test suite green (integration 41/41, other packages passing)
- **Commit shape**: 14 commits (spec + 4 feat + 1 fix + 4 docs + 1 review fix + 4 follow-ups) — intentionally preserved as separate commits for semantic-release changelog (both `feat` and `fix` types present)
- **Merge strategy**: **Merge commit only** — squashing would lose the changelog distinction

## Files changed

- `packages/core/src/types.ts` — add tierFloors, maxEmbedChars, WikiInvalidReadOptions
- `packages/core/src/readOptions.ts` — validateTierFloors, selectWithFloors
- `packages/core/src/services/RetrievalService.ts` — validation, both cut sites, metadata, ranker backfill reservation
- `packages/core/src/services/EmbeddingService.ts` — configurable clamp with ceiling
- `packages/core/src/utils/embedDefaults.ts` — constants export
- `packages/core/src/index.ts` — export constants
- `packages/core/__tests__/tierFloors.test.ts` — 11 validateTierFloors + selectWithFloors tests
- `packages/core/__tests__/tierFloorsRetrieval.test.ts` — 8 integration tests (ranker + fallback paths)
- `packages/core/__tests__/vectorRanker.test.ts` — 3 vectorRanker + tierFloors regression tests (review fix)
- `packages/core/__tests__/embedInputClamp.test.ts` — 4 embedFact clamp tests
- `packages/core/README.md` — document both options, update tierFloors contract language
- `README.md` — document both options, update tierFloors contract language
- `docs/superpowers/specs/2026-09-02-tier-floors-and-embed-clamp-design.md` — mark implemented, record deviations and #117 closure

## Closes

- Closes #109 (tierFloors)
- Closes #104 (embed clamp)
- Closes #117 (vectorRanker backfill gap)

https://claude.ai/code/session_013M4X7iJeN8ThKQhtLyEvcc


equationalapplications:
## /fix-pr follow-up

**Commit:** `7c14bc6`

### Review resolution

- **readOptions.ts: tierFloor non-finite key validation** — **Fixed.** Non-finite values are now filtered before key validation. `{ typo: NaN }` falls through cleanly. Finite floors keyed to unknown entities still throw. (files: `packages/core/src/readOptions.ts`, `packages/core/__tests__/tierFloors.test.ts`)
- **RetrievalService.ts: empty-query validation bypass** — **Fixed.** `validateTierFloors` now runs only when `trimmedQuery` is non-empty. Empty-query reads return recency results even with floors-exceed-maxResults configurations. Non-empty queries still validate. (files: `packages/core/src/services/RetrievalService.ts`, `packages/core/__tests__/tierFloorsRetrieval.test.ts`)
- **RetrievalService.ts: keyword-fallback starvation** — **Fixed.** Full keyword candidate set is materialized when any positive floor is active, mirroring the JS-cosine `jsCosineNeedsTierSort` widening. The oversampled `maxResults*2` cap is preserved when no floor is active. (files: `packages/core/src/services/RetrievalService.ts`, `packages/core/__tests__/tierFloorsRetrieval.test.ts`)
- **RetrievalService.ts: keyword-rank fallback preFilterLimit** — **Fixed.** When `effectivePreFilterLimit` is set, the keyword-rank fallback restricts its MiniSearch call to `candidateRows`’ IDs, so a floor cannot resurrect a fact the pre-filter excluded (§3.5). (files: `packages/core/src/services/RetrievalService.ts`, `packages/core/__tests__/vectorRanker.test.ts`)

### Verification

- typecheck — pass (`pnpm --filter @equationalapplications/core-llm-wiki typecheck`; full monorepo `pnpm typecheck` clean across 11 of 12 workspace projects)
- lint — N/A (no lint script in repo)
- tests — pass (`pnpm --filter @equationalapplications/core-llm-wiki test` → 1232/1232 passing, +6 new regression tests)

equationalapplications:
## /fix-pr follow-up

**Commits (newest first):**
- `1927b5e` docs(spec): record #117 closure in tierFloors + embed clamp spec Status
- `936c797` docs(core): describe tierFloors as reserving each namespace's top-N results
- `c844176` fix(core): close vectorRanker tierFloors gap (closes #117)

### Review resolution

- **RetrievalService.ts:414 (moderate, 2 votes) — external-ranker keyword-fallback oversampling floor-aware.** Fixed. The external-ranker keyword-fallback path now mirrors the JS-cosine path's `jsCosineNeedsTierSort` widening and the ordinary keyword-fallback path's `hasActiveFloors` widening: when any positive sanitized floor is active, the MiniSearch oversampling limit widens to `Number.MAX_SAFE_INTEGER`. Pre-filtered ID restriction is preserved. Regression test added in `vectorRanker.test.ts`: "keyword fallback honors a tier floor beyond the oversampling window when ranker fails" (60 `big` + 2 `small` facts, maxResults=5, tierFloors={small:2}).

- **RetrievalService.ts:486 (moderate, 1 vote) — reserve per-entity floor shortfalls during external-ranker backfill.** Fixed. Before computing the global backfill budget the ranker path now computes `floorShortfallByEntity` from the ranker output and pre-reserves that many omitted rows per floored entity (sorted by kwScore with the shared tie-break), then pushes them into `scored` (pure-semantic: score -2; hybrid: score (1-weight)*kwScore). Both branches of the backfill selection skip `reservedIds`. The pure-semantic global budget is now `max(0, maxResults - scored.length)` computed **after** reservation, so a floored entity whose ranker output was empty still satisfies its floor. Two regression tests added: pure-semantic "ranker omits an entire entity" and hybrid "ranker backfill reservation honors a floor in hybrid mode too".

- **README.md:32 (nit, 1 vote, suppressed) — describe floors as reserving top-N matching results, not guaranteeing a minimum contribution.** Fixed. README.md and packages/core/README.md now say "reserving each namespace's top-N matching results". Also drops the now-stale "Known limitation with an external `vectorRanker`" sub-bullet under `tierFloors` and replaces it with a one-line note that the ranker backfill path reserves per-entity floor shortfalls. The exception that an entity with fewer matching facts than its floor contributes what it has is unchanged — it's documented in the tierFloors definition block where the contract actually belongs.

The #117 follow-up gap is closed; per memory `feedback_separate-spec-and-code-commits` the spec Status was updated in a separate commit so the record shows the change with its date.

### Verification

- **typecheck** — pass (`pnpm typecheck`, 11/11 packages)
- **lint** — N/A (no lint script in repo)
- **tests** — pass (1421/1421 core, 41/41 integration; full monorepo `pnpm test` green)

equationalapplications:
## Self-review correction

**Commit:** `86fa686`

I reviewed my own fixes by reverting only `RetrievalService.ts` to `c844176^` and re-running the three tests I added. Two findings, both now fixed.

### 1. The hybrid regression test was not a regression test

`ranker backfill reservation honors a floor in hybrid mode too` **passed against the pre-fix code**. The corpus didn't reproduce the bug: the query was `kiwi`, which matched only the two `small` facts, so they held the only non-zero keyword scores and won the global backfill top-K unaided. Reservation was never exercised — the test would not have caught a regression.

Rebuilt so the hybrid starvation mechanism is actually present: 20 `big` facts match the query, the ranker scores only 5, leaving 15 keyword-matching `big` rows competing for a backfill budget of 5 against 2 `small` rows that don't match the query at all. Without reservation the top-K is 5 `big` rows and no `small` row reaches `selectWithFloors`.

Verified against `c844176^` (all three now fail pre-fix) and HEAD (all pass):

| Test | pre-fix | post-fix |
|---|---|---|
| keyword fallback honors a tier floor beyond the oversampling window | ✗ fail | ✓ pass |
| ranker backfill reservation … omits an entire entity (pure semantic) | ✗ fail | ✓ pass |
| ranker backfill reservation … hybrid mode | ✗ fail *(was ✓ pass)* | ✓ pass |

The other two were valid as written — I confirmed via a probe that MiniSearch does place the `small` facts at positions 60–61, outside the 55-row oversampling window, so that corpus genuinely starves.

### 2. `maxBackfill` didn't match its own comment

The comment read "Compute backfill budget AFTER reservation … pure semantic fills whatever slots remain after reserved rows have been pre-admitted", but the statement sat *above* the reserved-row push. Pure-semantic mode therefore sized the budget from a `scored.length` that excluded the reserved rows, over-materializing by `reserved.length`. Results were already correct — `selectWithFloors` truncates the surplus and floors were satisfied — so this is tidiness, not a behavior fix. Moved the computation below the push so the code does what it says.

### Correction to my earlier comment

I reported tests as "1421/1421 core". That figure came from a repo-root `vitest run` spanning multiple packages, not the core suite. The core package suite is **1235/1235**. The spec's Status line carried the same mislabel and is corrected in this commit.

### Verification

- typecheck — pass (`pnpm typecheck`, 11/11 packages)
- lint — N/A (no lint script in repo)
- tests — pass (`pnpm test`): core 1235/1235, integration 41/41, all other packages green

https://claude.ai/code/session_013M4X7iJeN8ThKQhtLyEvcc

PR #131, opened 2026-09-05
fix(scopelab): make the test suite self-sufficient on a fresh clone (#122)

### LINKED ISSUE #122: test(scopelab): suite fails on fresh clones — hidden dependency on built workspace dist output
## Summary

`pnpm --filter scopelab test` fails on a fresh clone with "Failed to resolve entry" until workspace packages are built first (`pnpm -r --filter './packages/*' build`). Scopelab's vitest resolves `core-llm-tools` (and possibly other workspace deps) through built `dist/` output rather than source.

Found during self-review of the authorized-scopes centralization work; the review had to build workspace packages in a throwaway worktree before the scopelab suite would even collect.

## Suggested fixes (either)

1. Add a `pretest` (or dependency) build step to scopelab's package script so test runs are self-sufficient.
2. Point scopelab's vitest config at workspace source via aliases, removing the dist dependency entirely.

Option 2 is preferable if aliasing stays simple — it also removes stale-dist false positives.

equationalapplications:
Implements §6 of the marker lifecycle & scopelab hygiene spec.

Scopelab had no vitest config, so vitest inherited `vite.config.ts` and
resolved `core-llm-tools` and `core-llm-wiki` through their `exports` maps
into `dist/`. Nothing builds those packages before the scopelab suite runs,
so a fresh clone failed with "Failed to resolve entry".

A dedicated `vitest.config.ts` aliases both workspace deps to their
TypeScript source. Chosen over a `pretest` build step because it removes the
dist coupling rather than papering over it, eliminates stale-dist false
positives (a suite passing against a `dist/` older than its `src/`), keeps
the feedback loop fast, and drops the PWA and wasm-copy plugins from the
test path where they do nothing.

Verified from a clean state: `rm -rf packages/*/dist && pnpm --filter
scopelab test` is green, and `pnpm --filter scopelab build` is unaffected.

No test content changed — only resolution.

Closes #122.

### DIFF
--- a/apps/scopelab/vitest.config.ts
+++ b/apps/scopelab/vitest.config.ts
@@ -1,0 +1,37 @@
+import { defineConfig } from 'vitest/config'
+import { resolve } from 'path'
+
+// Tests resolve workspace deps to SOURCE, not built dist output. Without this,
+// vitest inherits vite.config.ts, resolves these packages through their
+// `exports` maps into dist/, and the suite fails on a fresh clone until the
+// whole workspace is built (#122). Aliasing also removes stale-dist false
+// positives, where a suite passes against a dist/ older than its src/.
+//
+// NOTE: a dedicated vitest.config.ts REPLACES vite.config.ts for test runs
+// rather than merging with it. That intentionally drops the PWA and
+// sql-wasm-copy plugins, which are build concerns with no role in tests. Any
+// future test-relevant Vite setting must be repeated here.
+export default defineConfig({
+  resolve: {
+    alias: {
+      '@equationalapplications/core-llm-tools': resolve(
+        __dirname,
+        '../../packages/core-llm-tools/src/index.ts',
+      ),
+      '@equationalapplications/core-llm-wiki': resolve(
+        __dirname,
+        '../../packages/core/src/index.ts',
+      ),
+      // Transitive: packages/core's source imports core-okf, so without this
+      // alias the suite still resolves okf through its exports map into the
+      // missing dist/ once the first two aliases pull core in from source.
+      '@equationalapplications/core-okf': resolve(
+        __dirname,
+        '../../packages/okf/src/index.ts',
+      ),
+    },
+  },
+  test: {
+    environment: 'node',
+  },
+})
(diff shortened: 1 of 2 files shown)


equationalapplications:
## /fix-pr follow-up

**Commit:** `dd172a2`

### Review resolution

- **Add `@equationalapplications/core-okf` to the approved Vitest alias contract** (spec §6.2) — **Fixed.** This was the known gap between the approved design and the shipped config: the block listed two aliases, the implementation needs three. Added the alias, changed "both workspace deps" to "all required workspace packages", and stated the underlying rule the snippet was silently encoding — *alias the import closure, not the direct deps* — so a future copy of the block does not reintroduce the omission.

  Recorded the closure verification too: `packages/core/src` imports exactly one workspace package (`@equationalapplications/core-okf`) and `packages/okf/src` imports none, so three is both complete and minimal. File: `docs/superpowers/specs/2026-09-04-marker-lifecycle-and-scopelab-hygiene-design.md`.

- **Run the promotion reset in one transaction** (spec §2.3) — **Fixed, in #133** (`8d86af9`), where the code lives. All three writes now commit inside one `db.withTransactionAsync`, with a rollback test against a live SQLite adapter. Full reasoning in that PR's follow-up comment.

- **Validate the embedding dimension before clearing markers** (spec §4.1) — **Not applied.** The mechanism is described accurately: `ImportExportService` validates size, 4-byte alignment, and finiteness (:221-247) but not dimension, and the dimension reconciliation runs *after* the upserts (:495+), so a wrong-dimension blob does land and now also clears markers.

  Not applied because the fix proposed — reject the blob before the upsert — changes import semantics well beyond this PR, and the wrong-dimension blob was already being persisted and already being skipped by `skipExisting` before this change; the delta here is only that the stale marker no longer survives alongside it. The designed signal for that state is `embedding_dimension_mismatch`, which this path does set (:520-534), and promotion clears the markers deliberately per §2.3. Dimension-aware import validation is a real gap, but it is a pre-existing one and belongs in its own change.

### Verification
- typecheck — pass (`pnpm typecheck`)
- tests — pass (`pnpm --filter scopelab test` → 2 files, 14 tests)

Note: this commit is documentation only; the `vitest.config.ts` in this PR already carried all three aliases.

PR #132, opened 2026-09-05
refactor(core-llm-tools)!: migrate executor off deprecated injector helper (#123)

### LINKED ISSUE #123: refactor(core-llm-tools): migrate executor off deprecated buildAuthorizedSchemaArray
## Summary

Spec §4 of the 2026-09-04 embedding-failure & scope-hygiene design centralized the three executable scope-check sites onto `AUTHORIZED_SCOPES` / `isAuthorizedScope`, but deliberately left the scopelab executor consuming the deprecated `buildAuthorizedSchemaArray` (explicit non-goal, spec §6).

This is the follow-up so that deferral isn't lost: migrate the executor in `apps/scopelab/src/lib/llm/function-caller.ts` off the deprecated injector helper onto the shared constant/guard, then remove the deprecated helper if no other consumers remain.

## Context

The deprecation was preserved intentionally to keep PR-1's diff minimal; nothing depends on the deprecated path semantically — the executor already checks scopes via the centralized guard at the execution site.

equationalapplications:
Implements §5 of the marker lifecycle & scopelab hygiene spec, completing
the deferral left open by #127 (spec §6 of the previous design).

- `function-caller.ts` now uses `buildAuthorizedToolsArray`. Not a rename:
  the new helper returns the Gemini `tools[]` wrapper already built, so both
  the request body and the `advertisedNames` derivation changed shape.
- Deprecated `buildAuthorizedSchemaArray` deleted, along with its export.
  **This removes a function from the published API surface of
  `@equationalapplications/core-llm-tools`.**
- `AUTHORIZED_SCOPES` is now `Object.freeze`d — `as const` is compile-time
  only, so a consumer could `push()` and silently widen always-on
  authorization for every downstream check.
- The scope-sync execution test loops over all `AUTHORIZED_SCOPES` members
  instead of indexing `[0]`, so it keeps its meaning as the list grows.

Behavior-neutral for scopelab: the two helpers differ only in their handling
of `built_in` manifests, and scopelab registers none. The pre-existing tests
passed unchanged through the migration commit, which is the equivalence proof.

Closes #123.

### DIFF
--- a/apps/scopelab/src/lib/llm/__tests__/function-caller.test.ts
+++ b/apps/scopelab/src/lib/llm/__tests__/function-caller.test.ts
@@ -35,7 +35,7 @@ function geminiText(text: string) {
   return geminiResponse({ candidates: candidates })
 }
 
-/** A manifest that buildAuthorizedSchemaArray will actually advertise (has `schema`). */
+/** A manifest that buildAuthorizedToolsArray will actually advertise (has `schema`). */
 function makeTool(over: Partial<{ name: string; scope: string; schemaName: string }> = {}) {
   const name = over.name ?? 'search_memory'
   return {
@@ -232,7 +232,7 @@ describe('chatWithMemory — fail-closed scope authorization (M-2)', () => {
 
   it('rejects a tool whose name is not in the advertised schemas', async () => {
     fetchMock.mockResolvedValueOnce(geminiFunctionCall('search_memory', { query: 'q' }))
-    // Tool with scope `core` (so buildAuthorizedSchemaArray advertises it)
+    // Tool with scope `core` (so buildAuthorizedToolsArray advertises it)
     // but the schema name diverges from the manifest name. The manifest
     // name IS advertised, so the manifest-name check passes and the
     // advertised-schema check must be what rejects the invocation.
--- a/apps/scopelab/src/lib/llm/__tests__/scope-sync.test.ts
+++ b/apps/scopelab/src/lib/llm/__tests__/scope-sync.test.ts
@@ -56,27 +56,31 @@ describe('injector/executor scope parity (issue #106)', () => {
     expect(advertised).toHaveLength(AUTHORIZED_SCOPES.length)
   })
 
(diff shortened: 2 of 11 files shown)


equationalapplications:
## /fix-pr follow-up

**Commit:** `a9a8d6c` (unchanged — no code changes were warranted)

### Review resolution

Both findings were checked against the code and neither reproduces.

- **"Code assumes all tool entries have a `functionDeclarations` property"** (`function-caller.ts:76`, flagged critical) — **Not applied; the premise is inverted.** The code does not assume it, it tests for it: `authorizedTools.flatMap(entry => 'functionDeclarations' in entry ? entry.functionDeclarations.map(d => d.name) : [])`. The `: []` branch is exactly the built-in-tool case.

  Built-in entries being absent from `advertisedNames` is the intended behavior, not a defect. Per `types.ts:88`, a `BuiltInToolManifest` is "executed server-side, no client handler" — Gemini runs `google_search` itself and never returns it as a client `functionCall`. If one ever did appear, rejecting it is the correct fail-closed outcome, since there is no handler to dispatch to. The adjacent `tools.find(t => t.name === functionCall.name)` would reject it regardless: built-ins carry no `schema` and are not in the executable manifest set.

- **"Test loop uses `vi.stubGlobal` without isolation between iterations"** (`scope-sync.test.ts:56`, flagged major) — **Not applied; no concurrency exists.** Vitest runs `it` blocks within a file sequentially unless declared `.concurrent`, and these are not. Each iteration also restores the global in its own `finally` via `vi.unstubAllGlobals()`, so the stub cannot outlive its iteration even if a body throws. `AUTHORIZED_SCOPES` currently has one member, so the loop runs once today in any case — it is written as a loop so that growth is covered automatically, which is the point of the change.

### Verification
- typecheck — pass (`pnpm typecheck`)
- tests — pass (`pnpm --filter scopelab test` → 14 tests; `pnpm --filter @equationalapplications/core-llm-tools test` → 6 files, 51 tests)

Issue #129, opened 2026-09-04
core: embedding marker & reembed follow-ups from #125/#126 self-review


equationalapplications:
## Summary

Deferred findings from the pre-merge self-review round of #125 (embedding failure persistence markers) and #126 (runReembed retry orchestration). All were judged minor/inert at review time; none block those PRs. Collecting them here so they aren't lost.

## Checklist

- [x] **Stale marker survives an import that brings a valid blob** — `EntryRepository.upsertFact` (`packages/core/src/repositories/EntryRepository.ts`, the ON CONFLICT clause) sets `embedding_blob` from the imported row but leaves the marker columns untouched, so importing a bundle with a valid embedding over a locally-marked row yields a row with a valid embedding AND `embedding_failed_at IS NOT NULL`. Behaviorally inert today (default `runReembed` sweeps re-embed everything, self-healing the marker; classification gates on blob validity first), but contradictory as a diagnostic. Fix: add `embedding_failed_at = CASE WHEN excluded.embedding_blob IS NOT NULL THEN NULL ELSE embedding_failed_at END` (and companions) to the SET clause — and pin it with a test either way.
- [x] **Export `EmbeddingMarkerKind` from the core package root** — ~~`packages/core/src/index.ts` exports `EmbedFactResult` and `EmbedFailureKind` but not `EmbeddingMarkerKind`, so hosts annotating marker kinds can't name the narrow type. One-line, semver-safe (types-only, additive).~~ **No change needed — already exported.** The type is declared at `packages/core/src/types.ts:875` and `packages/core/src/index.ts:4` is `export * from './types'`, so it is already importable from the package root. See comment below.
- [x] **Test `storeEmbeddingDimension` failure as a `storage_error` source** — the storage_error test in `EmbeddingService.test.ts` only rejects `updateEmbeddingBlob`; the taxonomy names both DAO writes. One mock rejection pins it.
- [x] **Widen `findAllForReembed` return type** — it types its return `WikiFact & { embedding_blob?… }` while its `SELECT *` also returns the marker columns at runtime, forcing a `row as unknown as {…}` double cast in `MaintenanceService.runReembed`. A `ReembedCandidateRow` type (or widened return) makes the column dependency visible at the type level.
[...]

PR #133, opened 2026-09-05
feat(core): embedding marker lifecycle (#121, #129)

### LINKED ISSUE #121: feat(core): clear embedding failure markers when embedding provider or dimension changes
## Summary

Permanent embedding-failure exclusion (introduced by the reembed retry orchestration) survives a provider or embedding-dimension change. Once a fact reaches `MAX_EMBED_ATTEMPTS` (5) or a single `float32_overflow` under provider A, the marker is never reset — switching `llmProvider`, or promoting a new `embedding_dimension` via `reconcileEmbeddingDimension`, inherits the old provider's failures.

## Why it matters

Permanently-failed facts have **no embedding** — they are exactly the rows a new provider or dimension is most likely to fix, and they are invisible to retrieval. Today the only escape is `runReembed({ force: true })`, which callers are unlikely to discover.

## Suggested approaches (pick one)

1. **Reset on dimension promotion**: when `reconcileEmbeddingDimension` promotes a new `embedding_dimension`, clear `embedding_failed_at` / `embedding_failure_kind` and reset `embedding_attempts = 0` for affected rows as part of the full re-embed sweep it already triggers.
2. **Reset on provider change**: detect `llmProvider` identity changes and clear marker state similarly.
3. **Document-only**: document that provider swaps require `force: true`.

Option 1 is the most natural fit since dimension promotion already implies every stored vector is stale.

## Context

Found in self-review of the reembed-retry branch (`MaintenanceService.ts` `classifyReembedRow` interacting with the dimension-mismatch recovery path). The implementation matches the spec as written — this is a gap in the permanent-exclusion design, filed so the `force`-only escape hatch is a documented decision rather than a discovered limitation.

equationalapplications:
Implements §2, §4.1, §4.3 and §4.4 of the marker lifecycle spec.

- Markers are cleared on embedding-dimension promotion, including
  `float32_overflow` — terminal only because retrying is the same arithmetic
  on the same vector, which stops being true after a model change.
- A truthy non-function `embed` is now `no_provider` instead of
  `provider_error`, so host misconfiguration no longer burns
  MAX_EMBED_ATTEMPTS.
- An upsert carrying a valid blob clears stale markers, in both `upsert` and
  `upsertForImport`.
- `findAllForReembed` returns `ReembedCandidateRow`, removing two casts.

Closes #121. Ticks items 1, 3 and 4 of #129.

Residual documented on `runReembed`: a same-dimension provider swap still
needs `runReembed({ force: true })`.

### DIFF
--- a/docs/superpowers/specs/2026-09-04-marker-lifecycle-and-scopelab-hygiene-design.md
+++ b/docs/superpowers/specs/2026-09-04-marker-lifecycle-and-scopelab-hygiene-design.md
@@ -1,0 +1,447 @@
+# Embedding Marker Lifecycle & Scopelab Test Hygiene — Design
+
+**Status:** Approved, not yet implemented (2026-09-04)
+
+**Issues addressed:** #121 (markers survive provider/dimension change), #129
+(deferred #125/#126 self-review findings), #123 (executor off deprecated
+injector helper), #122 (scopelab suite needs built dist), #124 (index on
+`embedding_failed_at` — **closed without code**, §3)
+
+**Extends:** `2026-09-04-embedding-failure-and-scope-hygiene-design.md`
+(Status: Implemented), whose §3 shipped the marker system this spec now
+completes and whose §6 deferred the #123 migration as an explicit non-goal.
+
+---
+
+## 1. Why these five travel together
+
+Four of the five are follow-ups filed out of the same self-review round on
+2026-09-04 (PRs #125–#128); the fifth (#124) is a performance question raised
+in the same round. They share a deferral trail, not code. Bundling them in one
+spec keeps that trail auditable in one place; §7 splits them into independent
+PRs because their blast radii differ sharply — one changes host-observable
+embedding semantics, two touch only a demo app.
+
+Investigation (2026-09-04) verified every file:line claim below against `main`
+@ `2684121`. **Three investigation findings materially changed the shape of the
+work versus the issue text** — §3 (#124 has no query to optimize), §4.2 (#129's
+`EmbeddingMarkerKind` item is already done), and §4.3 (#129 names a method that
+does not exist; the real fix has two sites). Each is recorded where it applies.
+
+---
+
+## 2. #121 — Markers survive a provider or dimension change
+
+### 2.1 Problem
+
+Marker state is cleared in exactly one place: a successful
+`EntryRepository.updateEmbeddingBlob` (`packages/core/src/repositories/EntryRepository.ts:885-897`)
+sets `embedding_failed_at = NULL`, `embedding_failure_kind = NULL`,
+`embedding_attempts = 0`. Nothing else ever clears it.
+
+So a fact that reached `MAX_EMBED_ATTEMPTS` (5) or took a single
+`float32_overflow` under provider A stays permanently excluded under provider B.
+`classifyReembedRow` (`packages/core/src/services/MaintenanceService.ts:82-102`)
+returns `'permanent'` for those rows on every subsequent sweep. Permanently
+failed rows have **no embedding at all**, so they are invisible to vector
+retrieval — they are exactly the rows a new provider is most likely to fix.
+The only escape today is `runReembed({ force: true })`, which a host is
+unlikely to discover.
+
+### 2.2 Decision: reset on dimension promotion, plus a callability guard
+
+Approach chosen (issue option 1) over provider-identity tracking (option 2) and
+docs-only (option 3). Option 2 would need a host-supplied provider fingerprint
+in `WikiOptions` and new metadata bookkeeping, to cover only the
+same-dimension-swap case that `force: true` already handles. Not worth the
+machinery.
+
+**Investigation finding that makes option 1 viable.** It is not obvious that
+dimension promotion is even reachable while permanently-failed rows exist.
+It is: `countStaleEmbeddings` (`EntryRepository.ts:658-670`) counts a row stale
+only when it has a **wrong-dimension blob** or a legacy `embedding` value —
+
+```sql
+(embedding_blob IS NOT NULL AND (CAST(length(embedding_blob) AS INTEGER) / 4) != ?)
+OR (embedding_blob IS NULL AND embedding IS NOT NULL)
+```
+
+A permanently-failed row has `embedding_blob IS NULL` and `embedding IS NULL`,
+so it is **not** counted stale and does not block promotion. Promotion therefore
+fires, and the reset hung off it actually runs.
+
+### 2.3 Change
+
+`EmbeddingService.reconcileEmbeddingDimension`
+(`packages/core/src/services/EmbeddingService.ts:46-56`) currently promotes and
+clears the mismatch key. Add a marker reset to the promotion branch, so the
+three writes are one logical event:
+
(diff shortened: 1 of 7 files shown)


equationalapplications:
## /fix-pr follow-up

**Commit:** `8d86af9`

### Review resolution

- **Dimension promotion is not transactional** (`MaintenanceService.ts:412`, and the same root raised by CodeRabbit on #131 §2.3) — **Fixed.** `reconcileEmbeddingDimension` now wraps all three writes — `setMeta('embedding_dimension')`, `clearDimensionMismatch`, `clearEmbeddingFailureMarkers` — in a single `db.withTransactionAsync`. The failure that motivated it: if the marker clear failed *after* `clearDimensionMismatch` committed, no `embedding_dimension_mismatch` key remained to re-trigger promotion, so those rows stranded until someone ran `runReembed({ force: true })`.

  Verified safe to open a transaction at this point rather than threading a handle in: both call sites — `MaintenanceService.runReembed:415` and `ImportExportService:524` — invoke it *outside* their own transactions (ImportExport's `withTransactionAsync` opens at :139 and closes at :432), and the adapter rejects nesting (`types.ts:25`).

  Files: `packages/core/src/services/EmbeddingService.ts`, `packages/core/__tests__/embeddingFailureMarkers.test.ts` (new rollback test against a live SQLite adapter — injects a failure into the marker clear and asserts the mismatch key, the canonical dimension, and the markers all survive, then that the same promotion succeeds once the failure clears), `packages/core/__tests__/services/EmbeddingService.test.ts` (fake adapter gained a `withTransactionAsync` pass-through).

- **Concurrent upsert can clear a marker mid-sweep and cause immediate re-embedding** — **Not applied.** Correctly describes a real interleaving, but it is a pre-existing property of every marker write on this path, not something this PR introduces, and the cost is bounded: the row gets one extra embed attempt. Closing it means a sweep-wide lock across the write path, which is a spec-level decision about `JobManager` lock granularity rather than a change this PR can carry. Tracking as a follow-up.

- **Dimension promotion clears markers for rows currently being processed by the sweep** — **Not applied**, and the `runReembed` half of the scenario does not occur: `reconcileEmbeddingDimension` is called *after* the `for (const row of rows)` loop finishes (`MaintenanceService.ts:423-425`), not during it, so a promotion cannot reclassify rows the same sweep is iterating. The residual case is a concurrent `importDump` promoting while a sweep runs — same lock-granularity follow-up as above.

### Verification
- typecheck — pass (`pnpm typecheck`, all packages "Done")
- tests — pass (`pnpm --filter @equationalapplications/core-llm-wiki test` → 99 files, 1287 tests; `pnpm test` → integration 7 files, 41 tests)

equationalapplications:
## /fix-pr follow-up (round 2)

**Commit:** `e743bd9`

#131 and #132 are now merged; this branch test-merges clean into current `main`.

### Review resolution

- **Classify removal of the exported helper as semver-major** (spec:411) — **Fixed.** The split rationale called removing `buildAuthorizedSchemaArray` "a semver-minor concern". That is wrong, and the finding is right: it was a public export, so removing it breaks any consumer still importing it.

  Worth noting the *shipped* release was already correct — PR-B landed as `ff21a4a refactor(core-llm-tools)!: remove deprecated buildAuthorizedSchemaArray`, and the `!` makes semantic-release cut a major for the package. So this was a stale spec sentence, not a release defect. Corrected the text and recorded the mechanism that makes the major happen, so the two cannot drift apart again. File: `docs/superpowers/specs/2026-09-04-marker-lifecycle-and-scopelab-hygiene-design.md`.

- **Transaction failure leaves markers stranded without mismatch key** (`EmbeddingService.ts:56`) — **Not applied; already fixed in this PR.** This is the finding from round 1, and `8d86af9` addressed it: the three writes now run inside one `db.withTransactionAsync` (`EmbeddingService.ts:68-72`), so a failure anywhere in the promotion rolls all of it back and the mismatch key survives to re-trigger. `embeddingFailureMarkers.test.ts` pins exactly that rollback against a live SQLite adapter.

  The scenario as written also has the ordering reversed — it posits a failure "after `clearDimensionMismatch()` but before `setMeta()`", but `setMeta` is the *first* of the three calls, not the last.

- **`upsert()` changes contract by clearing markers on conflict** (`EntryRepository.ts:219`) — **Not applied; this is the feature, not a side effect.** Clearing embedding failure markers when a conflict carries a valid `embedding_blob` is the change #121 asks for, specified in §4.1 and documented in the doc comment directly above the method.

  The rationale: a row that now *has* a valid embedding but still carries a failure marker is a self-contradictory diagnostic, and it inflates `runReembed`'s `permanentlyFailed` counter for a row that is no longer failing. The write is deliberately narrow — the `CASE WHEN excluded.embedding_blob IS NOT NULL` guard means a conflict with no blob leaves marker state untouched, matching the "absent means don't touch" semantics already used for `embedding_blob` itself. A caller passing a blob is asserting a successful embed; the marker state going with it is the point.

### Verification
- typecheck — pass (`pnpm typecheck`, no errors)
- tests — pass (`pnpm --filter @equationalapplications/core-llm-wiki test` → 99 files, 1287 tests; `pnpm test` → integration 7 files, 41 tests)
- merge — clean against current `main` (verified by local test-merge, post #131/#132)

equationalapplications:
## /fix-pr follow-up (round 3)

**Commit:** `e743bd9` (unchanged — the re-review raised no new valid findings)

The automated reviewer re-posted its two previous findings against the new commit. Both were addressed in rounds 1 and 2; neither reproduces against the current code.

- **"Transaction failure leaves markers stranded without a mismatch key"** (`EmbeddingService.ts:56`) — **Not applied.** The scenario requires the promotion to "succeed in setting the new dimension and clearing the mismatch key, but fail to clear markers" — three writes with two outcomes. Since `8d86af9` they are one transaction (`EmbeddingService.ts:68-72`), so that split cannot occur: either all three commit or the whole promotion rolls back with the mismatch key intact. `embeddingFailureMarkers.test.ts` injects a failure into precisely the third write and asserts the first two rolled back.

- **"upsert changes contract by clearing markers without a clear caller update"** (`EntryRepository.ts:219`) — **Not applied; the stated failure is inverted.** The claim is that "a caller using `upsert` to update a fact's metadata *without* providing an `embedding_blob` will have its existing failure markers cleared." The guard does the opposite:

  ```sql
  embedding_failure_kind = CASE WHEN excluded.embedding_blob IS NOT NULL
                                THEN NULL ELSE embedding_failure_kind END
  ```

  No blob → the `ELSE` branch → the existing marker is written back unchanged. A blob-free metadata update cannot clear markers.

  This exact property is already pinned by a passing test — `embeddingFailureMarkers.test.ts:188`, *"upsert WITHOUT a blob leaves existing markers intact"* — which upserts a title/body change with `embedding_blob: null` against a marked row and asserts all three marker columns survive (`embedding_failed_at` 1000, kind `provider_error`, attempts 1).

### Verification
- typecheck — pass (`pnpm typecheck`, no errors)
- tests — pass (`pnpm --filter @equationalapplications/core-llm-wiki test` → 99 files, 1287 tests; `pnpm test` → integration 7 files, 41 tests)
- merge — clean against current `main`

PR #147, opened 2026-09-05
fix(core): defer embedding-dimension promotion during reembed sweeps

### LINKED ISSUE #134: core: marker clear during dimension promotion is not transactional with a concurrent sweep
## Summary

Flagged during review of #133 (aws-cloud-agent-pr-review, `MaintenanceService.ts:412`). Two related concurrency gaps, both pre-existing patterns this PR did not introduce, and neither blocking that merge.

## The gaps

1. **Concurrent upsert clears a marker mid-sweep.** A row can be classified `permanent` (e.g. `float32_overflow`) by `runReembed`'s per-row loop. A concurrent `upsert`/`upsertForImport` carrying a valid blob for the same row clears its markers between classification and the next sweep. The next sweep sees clean marker state and re-embeds a row the previous sweep had just excluded as permanently failed. Bounded cost: one extra embed attempt, not data loss.

2. **Dimension promotion during an active sweep.** `reconcileEmbeddingDimension` runs after `runReembed`'s per-row loop completes (`MaintenanceService.ts:423-425`), so it cannot reclassify rows the *same* sweep is iterating — that part of the original finding didn't reproduce. The residual case is a concurrent `importDump` triggering promotion (`ImportExportService.ts:524`) while a separate `runReembed` sweep is in flight against the same DB.

## Why not fixed in #133

Both are `JobManager` lock-granularity questions — `runReembed` holds a `reembed`/`global_reembed` lock (`MaintenanceService.ts` `acquireLock`), but `upsert`/`upsertForImport` and `reconcileEmbeddingDimension` don't take or check it. Closing this means deciding how wide that lock should be (row-level? whole-DB? per-entity against reembed only?), which is a spec-level decision, not a one-PR fix.

## Suggested next step

Decide the intended lock scope for embedding marker mutation vs. reembed sweeps, record it in a short design note, then implement.

Source: #133 review (aws-cloud-agent-pr-review), 2026-09-05.

equationalapplications:
Closes #134.

`reconcileEmbeddingDimension` clears embedding failure markers for every
entity, but its callers hold only entity-scoped locks. An `importDump` of
entity B could therefore erase classifications made by an in-flight
`runReembed('A')`, resurrecting rows that sweep had excluded.

**Decision (spec §2):** a marker mutation is legal during a sweep iff it
cannot resurrect a row the sweep classified. The one global mutation defers
instead of widening a lock.

- `JobManager.isAnyReembedActive()` — read-only predicate, no new lock types
- `importDump` gates on it and skips the promotion; the sticky
  `embedding_dimension_mismatch` key means the next sweep tail or import
  completes it
- `tryAcquireAutoHealLock` refuses during sweeps, making an accidental
  safety property structural

No public API change, no lock widening, no new `WikiBusyError` paths.

Spec: `docs/superpowers/specs/2026-09-05-reembed-lock-scope-design.md`

### DIFF
--- a/docs/superpowers/specs/2026-09-05-reembed-lock-scope-design.md
+++ b/docs/superpowers/specs/2026-09-05-reembed-lock-scope-design.md
@@ -1,0 +1,218 @@
+# Reembed Lock Scope: Defer-While-Sweeping Promotion — Design
+
+**Status:** Approved, not yet implemented (2026-09-05)
+
+**Issue addressed:** #134 (marker clear during dimension promotion is not
+transactional with a concurrent sweep)
+
+**Extends:** `2026-09-04-marker-lifecycle-and-scopelab-hygiene-design.md`
+(spec §2 — marker lifecycle), whose §2 shipped the promotion-clears-markers
+behavior this spec now makes concurrency-safe, and
+`2026-09-04-embedding-failure-and-scope-hygiene-design.md`, which shipped the
+marker system itself.
+
+---
+
+## 1. Problem and scope decision
+
+Issue #134 (filed from the #133 review) describes two concurrency gaps:
+
+1. A row classified `permanent` by `runReembed`'s per-row loop has its
+   markers cleared mid-sweep by a concurrent `upsert`/`upsertForImport`
+   carrying a valid blob; the next sweep re-embeds a row the previous sweep
+   had excluded. Bounded cost: one extra embed attempt.
+2. A concurrent `importDump` triggers dimension promotion
+   (`reconcileEmbeddingDimension`) while a separate `runReembed` sweep is in
+   flight against the same DB.
+
+The issue framed both as open lock-granularity questions ("row-level?
+whole-DB? per-entity against reembed only?") requiring a spec-level decision.
+This spec records that decision.
+
+### 1.1 Concurrency model (constraint agreed 2026-09-05)
+
+The `JobManager` is an in-process, in-memory lock table constructed per
+`WikiMemory` instance (`WikiMemory.ts:112`). Its existence as the sole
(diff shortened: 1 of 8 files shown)


equationalapplications:
## /fix-pr follow-up

**Commit:** `1755f0e`

### Review resolution
- **"Use the qualified sweep-tail guarantee" (spec §2/§3.2, Minor) — Fixed:** both claim sites now say "*productive* sweep tail", with the `embedded > 0` condition stated inline (MaintenanceService runs the tail reconciliation only when the sweep embedded rows). (file: `docs/superpowers/specs/2026-09-05-reembed-lock-scope-design.md`)
- **"Make promotion coordination atomic with sweep start" (ImportExportService.ts:532, Major) — Not applied:** the finding describes a TOCTOU that cannot occur. The gate at that line runs inside `doImportEntity`, while the **global import key is held** (`acquireImportLocks` acquires it before any per-entity work). While that key is held, `JobManager.acquireLock` rejects every operation except `global_import` — including `runReembed`'s sweep acquisition (`JobManager.ts` maintenance-switch: reembed keys are checked and rejected under `global_import`). So no sweep can start between the `isAnyReembedActive()` check and the promotion transaction; the snapshot is sufficient because the window is already serialized by the import lock. The spec records this argument explicitly (§3.2, "why the gate has no TOCTOU"), and the branch's final review independently verified the lock chain. Adding an atomic reservation operation would duplicate that existing serialization.

### Verification
- typecheck — N/A (markdown-only change; `tsc --noEmit` was clean on the preceding code commits)
- lint — N/A (no lint script configured for docs)
- tests — pass, 1298/1298 in 100 files (`cd packages/core && npm test`)

PR #188, opened 2026-09-17
fix(core): reject foreign-owned graph node writes

equationalapplications:
Spec: docs/superpowers/specs/2026-09-17-graph-entity-isolation-design.md (revision -d, `c3c66ed`).

Rejects writes whose node IDs are already owned by another entity at both the graph preflight and the repository write boundary, without disclosing the foreign owner, node ID, or stored content. Soft-deleted rows continue to reserve their IDs.

- New public domain error `WikiGraphNodeOwnershipConflict` (code `WIKI_GRAPH_NODE_OWNERSHIP_CONFLICT`, fixed message, no cause) exported from both public entrypoints.
- `EntryRepository.upsert` adds `entity_id` to its existing pre-write lookup, throws early for a foreign owner, makes the conflict update conditional on owner equality, and re-reads the stored owner only when an adapter reports zero changes (`upsertForImport` untouched).
- `upsertGraphCore` runs a repository-backed ownership preflight as its first statement, before ontology resolution or any persistent mutation.
- `SQLiteAdapter.runAsync` JSDoc now states the affected-row contract required across the codebase; `upsert`'s zero-count re-read is local defensive behavior, not permission for under-reporting.
- Both ingestion paths inherit enforcement; full-path rejection runs in the graph preflight, partial-path rejection runs in `EntryRepository.upsert` (verified by direct call-routing spies). Both rely on host transaction rollback for atomicity after the preflight boundary.

Tests cover runtime/privacy exports (2), real-SQLite repository/preflight suites (15), forced collisions on both ingestion paths with routing assertions (2), and the SQLite adapter contract (integration + Expo driver forwarding). Targeted full-core regression: 1,317/1,317 across 103 files; Expo 12/12; typecheck + build clean on both packages; `git diff --check` clean.


equationalapplications:
## /fix-pr follow-up

**Commits:**
- `d3463e5` docs(spec): close graph entity isolation implementation status
- `5e2e063` fix(core,docs): withTransactionAsync README examples + JSDoc coverage

### Review resolution

- **Thread (append implementation status revision):** `docs/superpowers/specs/2026-09-17-graph-entity-isolation-design.md` — **Fixed** in `d3463e5`. New revision `2026-09-17-e` records verification (build/typecheck/test counts, `git diff --check`), the pre-merge CI deltas, and the closure note. Historical revisions `2026-09-17-a` through `-d` are preserved verbatim.
- **Thread (transaction callback contract in adapter examples):** `packages/core/README.md` — **Fixed** in `5e2e063`. Interface signature now `withTransactionAsync<T>(fn: (tx: SQLiteAdapter) => Promise<T>)`; both example impls (sql.js line 1191, better-sqlite3 line 1220) now invoke `fn(this)` so the host callback receives the active tx handle. Verified against the actual signature in `packages/core/src/types.ts:39`.
- **Docstring Coverage CI failure:** **Fixed** in `5e2e063`. `WikiGraphNodeOwnershipConflict` constructor in `packages/core/src/types.ts` now carries a privacy-preserving JSDoc, satisfying the 80% threshold on the 5 touched functions.

### Verification

- build — pass (`pnpm --filter @equationalapplications/core-llm-wiki build`)
- core tests — pass (1,317/1,317 across 103 files)
- expo tests — pass (12/12)
- `git diff --check` — clean

